PR #70, opened 2018-02-16
Changed Option structures to use Builder pattern

mfateev:


src/main/java/com/uber/cadence/worker/WorkerOptions.java
```diff
@@ -17,112 +17,140 @@
...
+        DataConverter dataConverter = JsonDataConverter.getInstance();
+        int maxConcurrentActivityExecutionSize;
+        int maxWorkflowThreads;
+
+        /**
+         * When set to true doesn't poll on workflow task list even if there are registered workflows with a worker.
+         * For clarity prefer not registing workflow types with a {@link Worker} to setting this option.
+         * But it can be useful for disabling polling through configuration without a code change.
+         */
+        public Builder setDisableWorkflowWorker(boolean disableWorkflowWorker) {
```
venkat1109: is there a reason to take a boolean parameter to this function ? setDisableWorkflowWorker() itself is self documenting

src/main/java/com/uber/cadence/worker/WorkerOptions.java
```diff
@@ -17,112 +17,140 @@
...
+            this.disableWorkflowWorker = disableWorkflowWorker;
+            return this;
+        }
+
+        /**
+         * When set to true doesn't poll on activity task list even if there are registered activities with a worker.
+         * For clarity prefer not registing activity implementations with a {@link Worker} to setting this option.
+         * But it can be useful for disabling polling through configuration without a code change.
+         */
+        public Builder setDisableActivityWorker(boolean disableActivityWorker) {
```
venkat1109: same comment as above

src/main/java/com/uber/cadence/worker/WorkerOptions.java
```diff
@@ -17,112 +17,140 @@
...
+         */
+        public Builder setMaxConcurrentActivityExecutionSize(int maxConcurrentActivityExecutionSize) {
+            this.maxConcurrentActivityExecutionSize = maxConcurrentActivityExecutionSize;
+            return this;
+        }
+
+        /**
+         * Maximum size of a thread pool used to execute workflows.
+         */
+        public Builder setMaxWorkflowThreads(int maxWorkflowThreads) {
```
venkat1109: why expose MaxWorkflowThreads as a config option for Workflows but MaxConcurrentActivityExecutionSize for activities instead ?

PR #76, opened 2018-02-18
Review feedback

mfateev:


jontro:
Documentation on firstpage and in javadoc still states newClient instead of newInstance

PR #100, opened 2018-03-05
Worker refactoring

mfateev:
Refactored worker classes from abuse of inheritance to composition.


src/main/java/com/uber/cadence/internal/sync/SyncWorkflowWorker.java
```diff
@@ -20,129 +20,73 @@
...
-
-    private DecisionTaskPoller decisionTaskPoller;
-
-    private final DataConverter dataConverter;
-
-    public SyncWorkflowWorker(WorkflowService.Iface service, String domain, String taskListToPoll, DataConverter dataConverter) {
-        setIdentity(ManagementFactory.getRuntimeMXBean().getName());
-        workflowThreadPool = new ThreadPoolExecutor(1000, 1000,
+    public SyncWorkflowWorker(WorkflowService.Iface service, String domain, String taskList, SingleWorkerOptions options) {
+        ThreadPoolExecutor workflowThreadPool = new ThreadPoolExecutor(1000, 1000,
```
venkat1109: can you define a constant for THREAD_POOL_SIZE ?

src/main/java/com/uber/cadence/internal/worker/Poller.java
```diff
@@ -0,0 +1,206 @@
...
+                pollBackoffThrottler.success();
+            } catch (Throwable e) {
+                pollBackoffThrottler.failure();
+                if (!(e.getCause() instanceof InterruptedException)) {
+                    uncaughtExceptionHandler.uncaughtException(Thread.currentThread(), e);
+                }
+            } finally {
+                // Resubmit itself back to pollExecutor
+                if (!pollExecutor.isShutdown()) {
+                    pollExecutor.execute(this);
```
venkat1109: this is weird.. I am sure there is a reason for doing this (rather than a simple while loop)

src/main/java/com/uber/cadence/internal/worker/Poller.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+        if (options.getMaximumPollRatePerSecond() > 0.0) {
+            pollRateThrottler = new Throttler("poller", options.getMaximumPollRatePerSecond(),
+                    options.getMaximumPollRateIntervalMilliseconds());
+        }
+
+        pollExecutor = new ThreadPoolExecutor(options.getPollThreadCount(), options.getPollThreadCount(),
+                1, TimeUnit.SECONDS,
+                new LinkedBlockingQueue<>(options.getPollThreadCount()));
+        pollExecutor.setThreadFactory(new ExecutorThreadFactory(options.getPollThreadNamePrefix(),
```
venkat1109: see https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executors.html

src/main/java/com/uber/cadence/internal/worker/PollerOptions.java
```diff
@@ -0,0 +1,157 @@
...
+            this.maximumPollRateIntervalMilliseconds = maximumPollRateIntervalMilliseconds;
+            return this;
+        }
+
+        public Builder setMaximumPollRatePerSecond(double maximumPollRatePerSecond) {
+            this.maximumPollRatePerSecond = maximumPollRatePerSecond;
+            return this;
+        }
+
+        public Builder setPollBackoffCoefficient(double pollBackoffCoefficient) {
```
venkat1109: its not clear to me as to whether these options are for

(1) Poller backoff/retry on errors ? (or)
(2) Poller throttling ?

Since this is exposed to the users through Builder, might be worth clarifying or simplifying this

src/main/java/com/uber/cadence/worker/Worker.java
```diff
@@ -137,17 +146,9 @@ public Worker(WorkflowService.Iface service, String domain, String taskList, Wor
      * The reason for registration accepting workflow class, but not the workflow instance is
      * that workflows are stateful and a new instance is created for each workflow execution.
      * </p>
-     *
-     * @param workflowImplementationClasses
      */
     public void registerWorkflowImplementationTypes(Class<?>... workflowImplementationClasses) {
```
venkat1109: can you also document that register is a SET operation - it will overwrite previously registered types (assuming I understand this API right)

src/main/java/com/uber/cadence/worker/Worker.java
```diff
@@ -165,9 +166,7 @@ public void registerActivitiesImplementations(Object... activityImplementations)
             throw new IllegalStateException("disableActivityWorker is set in worker options");
         }
         checkStarted();
-        for (Object implementation : activityImplementations) {
-            activityWorker.addActivityImplementation(implementation);
-        }
+        activityWorker.setActivitiesImplementation(activityImplementations);
     }
 
     private void checkStarted() {
```
venkat1109: nit: this method should probably be called checkNotStarted()

PR #143, opened 2018-04-11
add a new workflow service constructor that takes sub channel as input

meiliang86:


src/main/java/com/uber/cadence/serviceclient/WorkflowServiceTChannel.java
```diff
@@ -355,6 +344,35 @@ public WorkflowServiceTChannel(String host, int port, ClientOptions options) {
...
+   * Creates Cadence client with specified sub channel and options.
+   *
+   * @param subChannel sub channel for communicating with cadence frontend service.
+   * @param options configuration options like rpc timeouts.
+   */
+  public WorkflowServiceTChannel(SubChannel subChannel, ClientOptions options) {
+    this.options = options;
+    this.thriftHeaders = getThriftHeaders();
+    // this.metricsReporter = new MetricsReporter(options.getMetricsClient());
+    this.tChannel = null;
```
mfateev: who's going to shutdown the tChannel in this case?

PR #145, opened 2018-04-12
Added List[Open|Closed]WorkflowExecutions to the TestWorkflowService

mfateev:


src/main/java/com/uber/cadence/internal/sync/TestWorkflowEnvironmentInternal.java
```diff
@@ -167,6 +172,9 @@ public String getDiagnostics() {
   @Override
   public void close() {
     for (Worker w : workers) {
+      if (!w.isStarted()) {
```
meiliang86: No need to shutdown in this case?

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -104,7 +104,7 @@
   }
 
   @Rule public TestName testName = new TestName();
-  @Rule public Timeout globalTimeout = Timeout.seconds(500);
+  @Rule public Timeout globalTimeout = Timeout.seconds(5);
```
meiliang86: Is this the wall clock timeout?

mfateev: Yes. It uses wall clock.

PR #146, opened 2018-04-13
Added thrift serialization to JsonDataConverter

mfateev:


src/main/java/com/uber/cadence/converter/JsonDataConverter.java
```diff
@@ -128,6 +146,16 @@ public JsonDataConverter(Function<GsonBuilder, GsonBuilder> builderInterceptor)
     if (content == null) {
       return null;
     }
+    // Deserialize thrift values.
```
meiliang86: Seems to be cleaner if we just have one try/catch block as above and put the switch logic inside.

PR #148, opened 2018-04-18
add metrics in java client

meiliang86:


src/main/java/com/uber/cadence/internal/metrics/NoopScope.java
```diff
@@ -0,0 +1,91 @@
...
+
+import com.uber.m3.tally.Buckets;
+import com.uber.m3.tally.Capabilities;
+import com.uber.m3.tally.RootScopeBuilder;
+import com.uber.m3.tally.Scope;
+import com.uber.m3.tally.StatsReporter;
+import com.uber.m3.util.Duration;
+import java.util.Map;
+
+public class NoopScope {
```
mfateev: +final

src/main/java/com/uber/cadence/internal/sync/POJOActivityTaskHandler.java
```diff
@@ -115,6 +126,8 @@ public void addActivityImplementation(Object activity) {
               timeoutException.getTimeoutType(),
               dataConverter.toData(timeoutException.getDetails()));
     }
+
+    metricsScope.counter(MetricsType.ACTIVITY_EXEC_FAILED_COUNTER).inc(1);
```
mfateev: Should we increment this one as well in case of Error?

src/main/java/com/uber/cadence/internal/sync/TestActivityEnvironmentInternal.java
```diff
@@ -173,7 +174,8 @@ public void registerActivitiesImplementations(Object... activityImplementations)
       task.setActivityType(new ActivityType().setName(activityType));
       IWorkflowService service = new WorkflowServiceWrapper(null);
       Result taskResult =
-          activityTaskHandler.handle(service, testEnvironmentOptions.getDomain(), task);
+          activityTaskHandler.handle(
+              service, testEnvironmentOptions.getDomain(), task, NoopScope.getInstance());
```
mfateev: I would add Metric scope to TestEnvironmentOptions. As users want to instrument their code as well they have to be able to inject their own implementation of the Scope for testing.

src/main/java/com/uber/cadence/internal/worker/ActivityWorker.java
```diff
@@ -40,7 +45,7 @@
 
   private static final Logger log = LoggerFactory.getLogger(ActivityWorker.class);
 
-  private static final String POLL_THREAD_NAME_PREFIX = "Activity Poller ";
+  private static final String POLL_THREAD_NAME_PREFIX = "Poller taskList=";
```
mfateev: Should say it it is activity or workflow one.

src/main/java/com/uber/cadence/internal/worker/ActivityWorker.java
```diff
@@ -142,7 +148,23 @@ public void resumePolling() {
...
 
-  private class TaskHandlerImpl implements PollTask.TaskHandler<PollForActivityTaskResponse> {
+  private class MeasurableActivityTask {
+    PollForActivityTaskResponse task;
+    Stopwatch sw;
+
+    MeasurableActivityTask(PollForActivityTaskResponse task, Stopwatch sw) {
+      Objects.requireNonNull(task);
+      Objects.requireNonNull(sw);
+      this.task = task;
```
mfateev: this.task = Objects.requireNotNull(task);

src/main/java/com/uber/cadence/internal/worker/ActivityWorker.java
```diff
@@ -152,55 +174,94 @@ private TaskHandlerImpl(ActivityTaskHandler handler) {
 
     @Override
     public void handle(
-        IWorkflowService service, String domain, String taskList, PollForActivityTaskResponse task)
+        IWorkflowService service, String domain, String taskList, MeasurableActivityTask task)
```
mfateev: I don't see why the interface change is needed. The stopwatch can be started and stopped in the scope of this method.

src/main/java/com/uber/cadence/internal/worker/ActivityWorker.java
```diff
@@ -152,55 +174,94 @@ private TaskHandlerImpl(ActivityTaskHandler handler) {
...
-        ActivityTaskHandler.Result response = handler.handle(service, domain, task);
-        sendReply(task, response);
+        Stopwatch sw = options.getMetricsScope().timer(MetricsType.ACTIVITY_EXEC_LATENCY).start();
+        ActivityTaskHandler.Result response =
+            handler.handle(service, domain, task.task, options.getMetricsScope());
+        sw.stop();
+
+        sw = options.getMetricsScope().timer(MetricsType.ACTIVITY_RESP_LATENCY).start();
+        sendReply(task.task, response);
+        sw.stop();
```
mfateev: Shouldn't we have a separate counter for the sending reply failures?

src/main/java/com/uber/cadence/internal/worker/WorkflowWorker.java
```diff
@@ -176,16 +182,26 @@ private TaskHandlerImpl(DecisionTaskHandler handler) {
...
         throws Exception {
+      Stopwatch sw =
+          options.getMetricsScope().timer(MetricsType.DECISION_EXECUTION_LATENCY).start();
       DecisionTaskHandler.Result response =
           handler.handleDecisionTask(new DecisionTaskWithHistoryIteratorImpl(task));
+      sw.stop();
+
+      sw = options.getMetricsScope().timer(MetricsType.DECISION_RESPONSE_LATENCY).start();
       sendReply(service, task.getTaskToken(), response);
+      sw.stop();
```
mfateev: What about failures responding?

src/main/java/com/uber/cadence/internal/external/GenericWorkflowClientExternalImpl.java
```diff
@@ -61,45 +70,54 @@ public IWorkflowService getService() {
...
-    //        }
-
-    StartWorkflowExecutionResponse result;
     try {
-      result = service.StartWorkflowExecution(request);
-    } catch (WorkflowExecutionAlreadyStartedError e) {
-      throw e;
-    } catch (TException e) {
-      throw CheckedExceptionWrapper.wrap(e);
+      StartWorkflowExecutionRequest request = new StartWorkflowExecutionRequest();
```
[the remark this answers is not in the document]
meiliang86: Done.

src/main/java/com/uber/cadence/internal/external/ManualActivityCompletionClientImpl.java
```diff
@@ -104,6 +111,7 @@ public void complete(Object result) {
       request.setRunID(execution.getRunId());
       try {
         service.RespondActivityTaskCompletedByID(request);
+        metricsScope.counter(MetricsType.ACTIVITY_TASK_COMPLETED_BY_ID_COUNTER).inc(1);
```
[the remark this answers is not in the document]
meiliang86: WorkflowService client has per request success/failure counts.

src/main/java/com/uber/cadence/internal/metrics/NoopScope.java
meiliang86: Yes.

src/main/java/com/uber/cadence/internal/sync/POJOActivityTaskHandler.java
meiliang86: I'm matching the go client behavior here. My understanding is that Error here is go's panic. Let me know if it's not true.

src/main/java/com/uber/cadence/internal/worker/ActivityWorker.java
meiliang86: When we use it we put the type in the end.
```java
            new PollerOptions.Builder(pollerOptions)
                .setPollThreadNamePrefix(
                    POLL_THREAD_NAME_PREFIX
                        + "\""
                        + taskList
                        + "\", domain=\""
                        + domain
                        + "\", type=\"activity\"")
                .build();
```

src/main/java/com/uber/cadence/internal/worker/ActivityWorker.java
meiliang86: This is to report e2e activity task latency. So the stopwatch is created at the beginning of the poll method and then stopped here.

src/main/java/com/uber/cadence/internal/worker/ActivityWorker.java
meiliang86: Done.

src/main/java/com/uber/cadence/internal/worker/ActivityWorker.java
meiliang86: client has it per each rpc call. If we think we need it here we can add another one.

src/main/java/com/uber/cadence/internal/worker/WorkflowWorker.java
meiliang86: Similar to above.

src/main/java/com/uber/cadence/internal/sync/TestActivityEnvironmentInternal.java
meiliang86: Good suggestion. I'll do it in the next changeset where I expose the metrics object in Workflow class.

src/main/java/com/uber/cadence/internal/worker/ActivityWorker.java
mfateev: I still think this change is ugly. Would you measure it directly in the PollTask.run method?

PR #149, opened 2018-04-20
expose metrics scope in workflow and de-dup metrics in replay mode

meiliang86:


src/test/java/com/uber/cadence/workflow/MetricsTest.java
```diff
@@ -0,0 +1,87 @@
...
+    @Override
+    public void execute() {
+      Workflow.getMetricsScope().counter("test-started").inc(1);
+
+      ChildWorkflowOptions options =
+          new ChildWorkflowOptions.Builder().setTaskList(taskList).build();
+      TestChildWorkflow workflow = Workflow.newChildWorkflowStub(TestChildWorkflow.class, options);
+      workflow.executeChild();
+
+      Workflow.getMetricsScope().counter("test-done").inc(1);
```
mfateev: Add a test of the workflow sleeping for some time and a metric measuring the latency of the sleep.

src/main/java/com/uber/cadence/internal/metrics/ReplayAwareScope.java
```diff
@@ -0,0 +1,183 @@
...
+      gauge.update(value);
+    }
+  }
+
+  private static class ReplayAwareTimer implements Timer {
+    Timer timer;
+    ReplayAware context;
+
+    public ReplayAwareTimer(Timer timer, ReplayAware context) {
+      this.timer = Objects.requireNonNull(timer);
```
mfateev: This is not going to work. The problem is that this implementation reuses the timer from the wrapped scope. And the TimerImpl.start calls into the system clock. So during replay the start time will be the time of the replay, not the time the start was called during the first execution. 
Make sure that Workflow.currentTimeMillis() is used everywhere when calculating latencies.

src/main/java/com/uber/cadence/internal/replay/DecisionContextImpl.java
```diff
@@ -40,6 +42,8 @@
 
   private final WorkflowContext workflowContext;
 
+  private Scope metricsScope;
```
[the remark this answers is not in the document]
meiliang86: I don't want to leak this pointer before constructor exits.

src/main/java/com/uber/cadence/workflow/Workflow.java
```diff
@@ -708,6 +709,10 @@ public static boolean isReplaying() {
     return WorkflowInternal.isReplaying();
   }
 
+  public static Scope getMetricsScope() {
```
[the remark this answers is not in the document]
meiliang86: Done.

src/test/java/com/uber/cadence/workflow/MetricsTest.java
meiliang86: Done.

src/main/java/com/uber/cadence/internal/metrics/ReplayAwareScope.java
meiliang86: Fixed.

src/test/java/com/uber/cadence/workflow/MetricsTest.java
```diff
@@ -83,5 +110,13 @@ public void testWorkflowMetrics() throws InterruptedException {
     verify(reporter, times(1)).reportCounter("test-done", null, 1);
     verify(reporter, times(1)).reportCounter("test-child-started", null, 1);
     verify(reporter, times(1)).reportCounter("test-child-done", null, 1);
+
```
mfateev: Why do you need that sleep above? 
In any case never use Thread.sleep in workflow unit tests. Use env.sleep instead.

meiliang86: This is for the metrics client to flush the cached values out. It's not related to workflow testing.

PR #153, opened 2018-04-27
implement replay aware logging

meiliang86:


src/main/java/com/uber/cadence/internal/worker/SingleWorkerOptions.java
```diff
@@ -43,6 +43,8 @@
 
     private Scope metricsScope;
 
+    private boolean enableLoggingInReplay = false;
```
mfateev: No need to specify value is it matches the default in Java. It is actually antipattern.

src/main/java/com/uber/cadence/workflow/Workflow.java
```diff
@@ -720,6 +723,32 @@ public static Scope getMetricsScope() {
...
+  /**
+   * Get logger to use inside workflow. Logs in replay mode are omitted unless {@param
+   * enableLoggingInReplay} is set to true in {@link WorkerOptions} when a worker starts up.
+   *
+   * @param name name to appear in logging.
+   * @return logger to use in workflow logic.
+   */
+  public static Logger getLogger(String name) {
+    Logger logger = LoggerFactory.getLogger(name);
+    return new ReplayAwareLogger(
```
mfateev: I would make ReplayAwareLogger non public and move this code to WorkflowInternal.

src/test/java/com/uber/cadence/workflow/LoggerTest.java
```diff
@@ -0,0 +1,136 @@
...
+        new WorkflowOptions.Builder()
+            .setExecutionStartToCloseTimeout(Duration.ofSeconds(1000))
+            .setTaskList(taskList)
+            .build();
+    LoggerTest.TestWorkflow workflow =
+        workflowClient.newWorkflowStub(LoggerTest.TestWorkflow.class, options);
+    workflow.execute(wfID);
+
+    String tempDir = System.getProperty("java.io.tmpdir");
+    System.out.println(tempDir);
```
mfateev: Remove println

src/test/java/com/uber/cadence/workflow/LoggerTest.java
```diff
@@ -0,0 +1,136 @@
...
+    File logFile = new File(tempDir, "log-cadence-java-client.log");
```
mfateev: I think you can avoid all this complexity of using a file and configuring it. 
Use http://oct.im/how-to-create-logback-loggers-dynamicallypragmatically.html
with https://github.com/qos-ch/logback/blob/master/logback-core/src/main/java/ch/qos/logback/core/read/ListAppender.java

src/main/java/com/uber/cadence/internal/logging/ReplayAwareLogger.java
```diff
@@ -0,0 +1,428 @@
+package com.uber.cadence.internal.logging;
+
+import com.uber.cadence.internal.replay.ReplayAware;
+import java.util.function.Supplier;
+import org.slf4j.Logger;
+import org.slf4j.Marker;
+
+public class ReplayAwareLogger implements Logger {
```
mfateev: Move ti implementation package and make non public

src/main/java/com/uber/cadence/internal/logging/ReplayAwareLogger.java
```diff
@@ -0,0 +1,428 @@
...
+  }
+
+  @Override
+  public String getName() {
+    return log.getName();
+  }
+
+  @Override
+  public boolean isTraceEnabled() {
+    return log.isTraceEnabled();
```
mfateev: This should include shouldSkipLogging as well as all similar Is...Enabled methods.

src/main/java/com/uber/cadence/internal/logging/ReplayAwareLogger.java
```diff
@@ -0,0 +1,428 @@
...
+
+  @Override
+  public void error(Marker marker, String msg, Throwable t) {
+    if (shouldSkipLogging()) return;
+
+    log.error(marker, msg, t);
+  }
+
+  private boolean shouldSkipLogging() {
+    return context.isReplaying() && !enableLoggingInReplay.get();
```
mfateev: I think it is incorrect as it doesn't log when not replaying. Should be
`!context.isReplaying() || (isReplaying && !enableLoggingInReplay())`
Make sure that unit test fails because of that before fixing.

meiliang86: if isReplaying is false then should skip is false as well, right?
The current unit test covers this logic. If it fails then we won't see anything in the log :)

src/main/java/com/uber/cadence/internal/worker/SingleWorkerOptions.java
meiliang86: Removed

src/test/java/com/uber/cadence/workflow/LoggerTest.java
meiliang86: Removed.

src/main/java/com/uber/cadence/internal/logging/ReplayAwareLogger.java
mfateev: If is replaying is false then it means that code is executed for the first time and should not skip logging

src/main/java/com/uber/cadence/internal/logging/ReplayAwareLogger.java
meiliang86: Done.

src/main/java/com/uber/cadence/internal/logging/ReplayAwareLogger.java
mfateev: Now I understand. I missed the shouldSkip is negative of shouldLog

src/test/java/com/uber/cadence/workflow/LoggerTest.java
```diff
@@ -0,0 +1,143 @@
...
+            .build();
+    LoggerTest.TestWorkflow workflow =
+        workflowClient.newWorkflowStub(LoggerTest.TestWorkflow.class, options);
+    workflow.execute(wfID);
+
+    assertEquals(
+        startLinesExpected, matchingLines(String.format("Start executing workflow %s.", wfID)));
+    assertEquals(1, matchingLines(String.format("Executing child workflow %s.", wfID)));
+    assertEquals(1, matchingLines(String.format("Done executing workflow %s.", wfID)));
+    listAppender.list.size();
```
mfateev: What does this line test?

meiliang86: This was from debugging. Removed.

PR #156, opened 2018-05-01
Preparation for enabling travis ci ...

meiliang86:
and fix an issue that cause test to hang if fork join pool only has single thread
It will be a followup task to run docker tests as well in CI.


build.gradle
```diff
@@ -60,17 +60,6 @@ compileJava {
             ".*/generated-sources/.*"
 }
 
-license {
```
mfateev: Whey are these removed?

src/main/java/com/uber/cadence/internal/testservice/TestWorkflowService.java
```diff
@@ -104,6 +104,8 @@
   // key->WorkflowId
   private final Map<WorkflowId, TestWorkflowMutableState> openExecutions = new HashMap<>();
 
+  private final ForkJoinPool forkJoinPool = new ForkJoinPool(4);
```
mfateev: Why 4? Can we make it unbounded?

.travis.yml
```diff
@@ -0,0 +1,15 @@
...
+  - SKIP_DOCKER_SERVICE=true # used to skip docker service tests in WorkflowTest for now.
+
+before_install:
+  - pushd /tmp
+  - wget http://www-us.apache.org/dist/thrift/0.9.3/thrift-0.9.3.tar.gz
+  - tar xfz thrift-0.9.3.tar.gz
+  - cd thrift-0.9.3
+  - ./configure --enable-libs=no --enable-tests=no --enable-tutorial=no --with-cpp=no --with-c_glib=no --with-java=yes --with-ruby=no --with-erlang=no --with-go=no --with-nodejs=no
+  - make -j2 && sudo make install
+  - popd
```
mfateev: new line at the end
will it provide code coverage?

build.gradle
meiliang86: These two sections are currently duplicated...See above.
Probably we had a bad merge previously.

mfateev: I see.

src/main/java/com/uber/cadence/internal/testservice/TestWorkflowService.java
mfateev: I think more than 1 should be OK. Let's keep it 4 for now.

PR #161, opened 2018-05-10
Added Workflow.mutableSideEffect

mfateev:


src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -2814,6 +2817,51 @@ public void testSideEffect() {
     tracer.setExpected("sideEffect", "sleep PT1S", "executeActivity customActivity1");
   }
 
+  private static final Map<String, Long> mutableSideEffectValue =
```
meiliang86: Seems using a queue will make this much simpler? Then we can get rid of all the sleep stuffs.

meiliang86: So i tried this idea. The interesting thing is that it only works if I retain the sleep in the workflow code. If that is removed then handleMutableSideEffectMarker will not be called and mutableSideEffect will always think the value does not exist.

mfateev: Interesting. It looks like it indicates that the current implementation is broken when the mutableSideEffect is called more than once per decision. Let me investigate.

src/main/java/com/uber/cadence/internal/replay/ClockDecisionContext.java
```diff
@@ -154,11 +187,91 @@ private void timerCancelled(long startEventId, Exception reason) {
...
+  }
+
   void handleMarkerRecorded(HistoryEvent event) {
     MarkerRecordedEventAttributes attributes = event.getMarkerRecordedEventAttributes();
-    if (SIDE_EFFECT_MARKER_NAME.equals(attributes.getMarkerName())) {
+    String name = attributes.getMarkerName();
+    if (SIDE_EFFECT_MARKER_NAME.equals(name)) {
       sideEffectResults.put(event.getEventId(), attributes.getDetails());
-    } else {
       log.warn("Unexpected marker: " + event);
```
meiliang86: This warning log should not be in the if branch. Will need another one.

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -2814,6 +2819,54 @@ public void testSideEffect() {
...
+    values.add(1234L);
+    values.add(123L); // expected to be ignored as it is smaller than 1234.
+    values.add(3456L);
+    mutableSideEffectValue.put(taskList, values);
+    CompletableFuture<String> result = WorkflowClient.execute(workflowStub::execute, taskList);
+    if (useExternalService) {
+      Thread.sleep(1000);
+    } else {
+      testEnvironment.sleep(Duration.ofSeconds(1));
+    }
```
meiliang86: These sleeps seems unnecessary?

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -2814,6 +2819,54 @@ public void testSideEffect() {
...
+                "id1",
+                Long.class,
+                (o, n) -> n > o,
+                () -> mutableSideEffectValue.get(taskList).poll());
+        if (result.length() > 0) {
+          result.append(", ");
+        }
+        result.append(value);
+        if (i >= 3) {
+          Workflow.sleep(Duration.ofSeconds(1));
```
meiliang86: Can you add a comment to say that this sleep is to test the replay mode logic? It's not obvious from the code.

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -2814,6 +2819,54 @@ public void testSideEffect() {
     tracer.setExpected("sideEffect", "sleep PT1S", "executeActivity customActivity1");
   }
 
+  private static final Map<String, Queue<Long>> mutableSideEffectValue =
```
meiliang86: I don't think we will need the dictionary. Single queue is enough?

src/main/java/com/uber/cadence/internal/replay/ClockDecisionContext.java
```diff
@@ -154,11 +187,93 @@ private void timerCancelled(long startEventId, Exception reason) {
...
+    MutableSideEffectResult result = mutableSideEffectResults.get(id);
+    Optional<byte[]> stored;
+    if (result == null) {
+      stored = Optional.empty();
+    } else {
+      stored = Optional.of(result.getData());
+    }
+    long eventId = decisions.getNextDecisionEventId();
+    int accessCount = result == null ? 0 : result.getAccessCount();
+    try {
```
meiliang86: This try/catch is now not necessary. The internal code does not throw checked exception.

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -2929,6 +2982,12 @@ public void await(String reason, Supplier<Boolean> unblockCondition) {
       return next.sideEffect(resultType, func);
     }
 
+    @Override
+    public <R> R mutableSideEffect(
+        String id, Class<R> returnType, BiPredicate<R, R> updated, Func<R> func) {
```
meiliang86: Add trace here.

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
mfateev: It is because the tests are executed in parallel. We need to fix all the other places where globals are used by workflow code.

PR #162, opened 2018-05-11
fix bugs that cause time skipping to not work in certain cases

meiliang86:


src/main/java/com/uber/cadence/internal/testservice/TestWorkflowStore.java
```diff
@@ -137,6 +137,8 @@ public PollForActivityTaskResponse getTask() {
 
   long save(RequestContext requestContext) throws InternalServiceError, EntityNotExistsError;
 
+  void applyTimerAndLocks(RequestContext ctx);
```
mfateev: may be applyTimersAndLocks

src/main/java/com/uber/cadence/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -133,7 +133,7 @@ void apply(RequestContext ctx)
   private final Map<String, StateMachine<TimerData>> timers = new HashMap<>();
   private final Map<String, StateMachine<SignalExternalData>> externalSignals = new HashMap<>();
   private StateMachine<WorkflowData> workflow;
-  private StateMachine<DecisionTaskData> decision;
+  private volatile StateMachine<DecisionTaskData> decision;
```
mfateev: May be reuse existing lock?

meiliang86: The existing lock seems to protect this. I'll revert this change.

src/main/java/com/uber/cadence/internal/testservice/TestWorkflowStore.java
meiliang86: Sure.

PR #165, opened 2018-05-15
self advancing timer refactor

meiliang86:
refactor timer locks to provide more diagnostic information
fix bugs


src/main/java/com/uber/cadence/internal/testservice/SelfAdvancingTimerImpl.java
```diff
@@ -128,6 +186,9 @@ private void runLocked() {
   private long systemTimeLastLocked = -1;
   private boolean emptyQueue = true;
 
+  @SuppressWarnings("JdkObsolete")
+  private final LinkedList<LockEvent> lockEvents = new LinkedList<>();
```
mfateev: Consider using set for faster removal

PR #166, opened 2018-05-16
implement getVersion

meiliang86:


src/main/java/com/uber/cadence/workflow/Workflow.java
```diff
@@ -798,6 +799,100 @@ public static boolean isReplaying() {
...
+   *   result = testActivities.activity3();
+   * }
+   * </code></pre>
+   *
+   * It is recommended to keep the GetVersion() call even if single branch is left:
+   *
+   * <pre><code>
+   * Workflow.getVersion("fooChange", 2, 2);
+   * String result;
+   * if (version == 1) {
```
mfateev: The whole if else should be just
result = testActivities.activity3();

src/main/java/com/uber/cadence/workflow/Workflow.java
```diff
@@ -798,6 +799,100 @@ public static boolean isReplaying() {
...
+   * for example change activity3 to activity4, you just need to update the maxVersion from 2 to 3.
+   *
+   * <p>Note that, you only need to preserve the first call to GetVersion() for each changeID. All
+   * subsequent call to GetVersion() with same changeID are safe to remove. However, if you really
+   * want to get rid of the first GetVersion() call as well, you can do so, but you need to make
+   * sure: 1) all older version executions are completed; 2) you can no longer use “fooChange” as
+   * changeID. If you ever need to make changes to that same part, you would need to use a different
+   * changeID like “fooChange-fix2”, and start minVersion from DefaultVersion again.
+   *
+   * @param changeID identifier of a particular change
```
mfateev: All calls to getVersion that share a changeID are guaranteed to return the same version number. Use this to perform multiple coordinated changes that should be enabled together.

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -107,7 +109,7 @@
    * When set to true increases test, activity and workflow timeouts to large values to support
    * stepping through code in a debugger without timing out.
    */
-  private static final boolean DEBUGGER_TIMEOUTS = false;
+  private static final boolean DEBUGGER_TIMEOUTS = true;
```
mfateev: revert

src/main/java/com/uber/cadence/internal/replay/MutableSideEffectHandler.java
```diff
@@ -0,0 +1,143 @@
...
+import com.uber.cadence.EventType;
+import com.uber.cadence.HistoryEvent;
+import com.uber.cadence.MarkerRecordedEventAttributes;
+import com.uber.cadence.internal.replay.DecisionContext.MutableSideEffectData;
+import com.uber.cadence.workflow.Functions.Func1;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+public class MutableSideEffectHandler {
```
mfateev: To me it looks like two different classes as they don't share logic 100%. If you believe that they are indeed the same then it should be called something else.

src/main/java/com/uber/cadence/internal/replay/ClockDecisionContext.java
```diff
@@ -271,4 +191,37 @@ void handleMarkerRecorded(HistoryEvent event) {
...
+    Optional<byte[]> result =
+        versionHandler.handle(
+            changeId,
+            markerDataSerializer,
+            markerDataDeserializer,
+            (stored) -> {
+              if (stored.isPresent()) {
+                return Optional.empty();
+              }
+              return Optional.of(ByteBuffer.allocate(4).putInt(maxSupported).array());
```
mfateev: Why are we using binary serialization here? Let's use DataConverter through a lambda passed here.

src/main/java/com/uber/cadence/internal/replay/ClockDecisionContext.java
```diff
@@ -271,4 +191,37 @@ void handleMarkerRecorded(HistoryEvent event) {
...
+            });
+
+    if (!result.isPresent()) {
+      return WorkflowInternal.DEFAULT_VERSION;
+    }
+    return ByteBuffer.wrap(result.get()).getInt();
+  }
+
+  private void validateVersion(String changeID, int version, int minSupported, int maxSupported) {
```
mfateev: I don't see it being used anywhere.

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -2863,6 +2865,78 @@ public void testMutableSideEffect() throws ExecutionException, InterruptedExcept
...
+      // Test adding a version check in non-replay code.
+      int version = Workflow.getVersion("test_change", Workflow.DEFAULT_VERSION, 1);
+      String result = "";
+      if (version == Workflow.DEFAULT_VERSION) {
+        result += testActivities.activity1("activity1");
+      } else {
+        result += testActivities.activity2("activity2", 2); // This is executed.
+      }
+
+      // Test version change in non-replay code.
```
mfateev: I would make range from Default to 2

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -2863,6 +2865,78 @@ public void testMutableSideEffect() throws ExecutionException, InterruptedExcept
...
+      version = Workflow.getVersion("test_change", 1, 2);
+      if (version == 1) {
+        result += testActivities.activity1("activity1"); // This is executed.
+      } else {
+        result += testActivities.activity2("activity2", 2);
+      }
+
+      // Test adding a version check in replay code.
```
mfateev: Add similar one for removal of getVersion call.

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -2863,6 +2865,78 @@ public void testMutableSideEffect() throws ExecutionException, InterruptedExcept
     assertEquals("1234, 1234, 1234, 3456", result);
   }
 
+  private static final Set<String> getVersionExecuted =
+      Collections.synchronizedSet(new HashSet<>());
+
+  public static class TestGetVersionWorkflowImpl implements TestWorkflow1 {
+
+    @Override
+    public String execute(String taskList) {
```
mfateev: Add test for getVersion min/max violation

src/main/java/com/uber/cadence/internal/replay/ActivityDecisionContext.java
```diff
@@ -137,6 +137,11 @@ void handleActivityTaskCompleted(HistoryEvent event) {
         byte[] result = attributes.getResult();
         BiConsumer<byte[], Exception> completionHandle = scheduled.getCompletionCallback();
         completionHandle.accept(result, null);
+      } else {
+        throw new Error(
```
mfateev: This is expected to be thrown on non deterministic change. Do you think it is more useful for the end user than the generic determinism error? Could we set this one as a cause to the generic one?

src/main/java/com/uber/cadence/internal/replay/DecisionsHelper.java
```diff
@@ -217,6 +230,8 @@ boolean handleStartChildWorkflowExecutionFailed(HistoryEvent event) {
    */
   long requestCancelExternalWorkflowExecution(
       RequestCancelExternalWorkflowExecutionDecisionAttributes schedule) {
+    addAllMissingVersionMarker(false, Optional.empty());
```
mfateev: Can   addAllMissingVersionMarker call be just part of getNextDecisionEventId()?

src/main/java/com/uber/cadence/internal/replay/DecisionsHelper.java
```diff
@@ -588,6 +620,57 @@ private void addDecision(DecisionId decisionId, DecisionStateMachine decision) {
...
+        .getMarkerRecordedEventAttributes()
+        .getMarkerName()
+        .equals(ClockDecisionContext.VERSION_MARKER_NAME)) {
+      return false;
+    }
+
+    // If we have a version marker in history event but not in decisions, let's add one.
+    if (!isNextDecisionVersionMarker
+        || (changeIdEquals.isPresent()
+            && changeIdEquals.get().test(event.getMarkerRecordedEventAttributes().getDetails()))) {
```
mfateev: Make event.getMarkerRecordedEventAttributes() a local variable as it is called multiple times within this method

src/main/java/com/uber/cadence/internal/replay/DecisionsHelper.java
```diff
@@ -588,6 +620,57 @@ private void addDecision(DecisionId decisionId, DecisionStateMachine decision) {
...
+
+    if (!event
+        .getMarkerRecordedEventAttributes()
+        .getMarkerName()
+        .equals(ClockDecisionContext.VERSION_MARKER_NAME)) {
+      return false;
+    }
+
+    // If we have a version marker in history event but not in decisions, let's add one.
+    if (!isNextDecisionVersionMarker
```
mfateev: Revert condition and return false from it.

src/main/java/com/uber/cadence/internal/replay/HistoryHelper.java
```diff
@@ -82,14 +83,23 @@
...
       }
       return decisionEvents.get(index);
     }
 
+    Optional<HistoryEvent> getOptionalDecisionEvent(long eventId) {
+      try {
+        HistoryEvent event = getDecisionEvent(eventId);
+        return Optional.of(event);
+      } catch (Exception e) {
+        return Optional.empty();
```
mfateev: Never use exceptions for flow control.
Either duplicate the code or implement getDecisionEvent in terms of getOptionalDecisionEvent.

src/main/java/com/uber/cadence/internal/replay/MutableMarkerHandler.java
```diff
@@ -0,0 +1,163 @@
...
+import com.uber.cadence.EventType;
+import com.uber.cadence.HistoryEvent;
+import com.uber.cadence.MarkerRecordedEventAttributes;
+import com.uber.cadence.converter.DataConverter;
+import com.uber.cadence.workflow.Functions.Func1;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+class MutableMarkerHandler {
```
mfateev: Rename as it is used not only for MutableSideEffectMarker

src/main/java/com/uber/cadence/internal/replay/MutableMarkerHandler.java
```diff
@@ -0,0 +1,163 @@
...
+      accessCount++;
+      return data;
+    }
+
+    int getAccessCount() {
+      return accessCount;
+    }
+  }
+
+  static final class MutableMarkerData {
```
mfateev: Rename

src/main/java/com/uber/cadence/internal/replay/DecisionsHelper.java
```diff
@@ -252,6 +267,10 @@ void handleRequestCancelExternalWorkflowExecutionFailed(HistoryEvent event) {
   }
 
   long signalExternalWorkflowExecution(SignalExternalWorkflowExecutionDecisionAttributes signal) {
+    addAllMissingVersionMarker(false, Optional.empty());
+
+    signal.setControl(getNextId().getBytes(StandardCharsets.UTF_8));
```
mfateev: Why do we need it back here?

src/main/java/com/uber/cadence/internal/replay/WorkflowDecisionContext.java
```diff
@@ -96,11 +95,7 @@ public void accept(Exception cause) {
     final StartChildWorkflowExecutionDecisionAttributes attributes =
         new StartChildWorkflowExecutionDecisionAttributes();
     attributes.setWorkflowType(parameters.getWorkflowType());
-    String workflowId = parameters.getWorkflowId();
-    if (workflowId == null) {
-      workflowId = generateUniqueId();
-    }
-    attributes.setWorkflowId(workflowId);
```
mfateev: Why is it removed? Is not server requiring it anymore?

src/main/java/com/uber/cadence/internal/sync/WorkflowInternal.java
```diff
@@ -58,6 +58,7 @@
  * Never reference directly. It is public only because Java doesn't have internal package support.
  */
 public final class WorkflowInternal {
+  public static final int DEFAULT_VERSION = -1;
```
mfateev: let's make it 0

src/main/java/com/uber/cadence/internal/replay/ActivityDecisionContext.java
meiliang86: I'll change it to throw NonDeterminisicWorkflowError.

src/main/java/com/uber/cadence/internal/replay/DecisionsHelper.java
meiliang86: addAllMissingVersionMarker is called from all the places that leads to addDecision, not just getNextDecisionEventId.

src/main/java/com/uber/cadence/internal/replay/MutableMarkerHandler.java
meiliang86: Done.

src/main/java/com/uber/cadence/internal/replay/MutableMarkerHandler.java
meiliang86: Done.

src/main/java/com/uber/cadence/internal/replay/DecisionsHelper.java
meiliang86: This needs to be done after addAllMissingVersionMarker. It used to be outside of this method and it's moved here..

src/main/java/com/uber/cadence/internal/replay/WorkflowDecisionContext.java
meiliang86: The same reason as above. But this will conflict with your new change and I'll just leave it as is.

src/main/java/com/uber/cadence/internal/sync/WorkflowInternal.java
meiliang86: Go uses -1 as default version so let's use the same here.

mfateev: We are Java ;)

PR #171, opened 2018-05-30
Deterministic retry

mfateev:


src/main/java/com/uber/cadence/internal/sync/WorkflowRetryerInternal.java
```diff
@@ -64,19 +65,24 @@ public static void retry(RetryOptions options, Functions.Proc proc) {
...
     options.validate();
     int attempt = 1;
-    long startTime = Workflow.currentTimeMillis();
+    long startTime = WorkflowInternal.currentTimeMillis();
+    // Records retry options in the history allowing changing them without breaking determinism.
+    String retryId = WorkflowInternal.randomUUID().toString();
     while (true) {
-      long nextSleepTime = calculateSleepTime(attempt, options);
+      RetryOptions retryOptions =
+          WorkflowInternal.mutableSideEffect(
```
meiliang86: Seems like this can be moved out of the while loop.

src/main/java/com/uber/cadence/internal/sync/WorkflowRetryerInternal.java
```diff
@@ -88,14 +94,19 @@ public static void retry(RetryOptions options, Functions.Proc proc) {
...
   private static <R> Promise<R> retryAsync(
-      RetryOptions options, Functions.Func<Promise<R>> func, long startTime, long attempt) {
+      String retryId,
+      RetryOptions options,
+      Functions.Func<Promise<R>> func,
+      long startTime,
+      long attempt) {
     options.validate();
-    CompletablePromise<R> funcResult = Workflow.newPromise();
+    CompletablePromise<R> funcResult = WorkflowInternal.newCompletablePromise();
```
meiliang86: We need to call WorkflowInternal.mutableSideEffect here to record retry option on the first (non-replay) attempt.

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -378,6 +378,55 @@ public void testActivityRetry() {
...
+    TestWorkflow1 workflowStub =
+        workflowClient.newWorkflowStub(
+            TestWorkflow1.class, newWorkflowOptionsBuilder(taskList).build());
+    try {
+      workflowStub.execute(taskList);
+      fail("unreachable");
+    } catch (WorkflowException e) {
+      assertTrue(e.getCause().getCause() instanceof IOException);
+    }
+    assertEquals(activitiesImpl.toString(), 3, activitiesImpl.invocations.size());
```
meiliang86: This assertion seems wrong. MaxAttempt is set to 2 in the non-replay path and should be honored here.
Same below.

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -1470,6 +1570,75 @@ public void testAsyncRetry() {
...
+
+    @Override
+    public List<String> getTrace() {
+      return trace;
+    }
+  }
+
+  /** @see DeterministicRunnerTest#testRetry() */
+  @Test
+  public void testAsyncRetryOptionsChange() {
```
meiliang86: Both of this and the sync test exercises the WorkflowRetryerInternal.retryAsync path. Should add a test for WorkflowRetryerInternal.retry as well.

PR #173, opened 2018-05-31
Activity heartbeat throttling

meiliang86:


src/main/java/com/uber/cadence/internal/sync/ActivityExecutionContextImpl.java
```diff
@@ -45,28 +49,88 @@
...
+  }
+
+  private void doHeartBeat(Object details) {
+    try {
+      sendHeartbeatDetails(details);
+      onHeartbeatSucceed();
+    } catch (TException e) {
+      // Not rethrowing to not fail activity implementation on intermittent connection or Cadence
+      // errors.
+      log.warn(
```
mfateev: Shouldn't activity and workflow ids be part of log context already?

src/main/java/com/uber/cadence/internal/sync/ActivityExecutionContextImpl.java
```diff
@@ -45,28 +49,88 @@
...
+      lock.unlock();
+    }
+  }
+
+  private void doHeartBeat(Object details) {
+    try {
+      sendHeartbeatDetails(details);
+      onHeartbeatSucceed();
+    } catch (TException e) {
+      // Not rethrowing to not fail activity implementation on intermittent connection or Cadence
```
mfateev: Shouldn't it keep retrying in case of intermittent errors even if recordActivityHeartbeat is not called again?

src/main/java/com/uber/cadence/internal/sync/ActivityExecutionContextImpl.java
```diff
@@ -45,28 +49,88 @@
...
+              } finally {
+                lock.unlock();
+              }
+            },
+            heartbeatIntervalNanos,
+            TimeUnit.NANOSECONDS);
+  }
+
+  private void sendHeartbeatDetails(Object details) throws TException {
     // TODO: call service with the specified minimal interval (through @ActivityExecutionOptions)
```
mfateev: remove TODO

src/main/java/com/uber/cadence/internal/sync/ActivityExecutionContextImpl.java
```diff
@@ -45,28 +49,88 @@
...
+                }
+              } finally {
+                lock.unlock();
+              }
+            },
+            heartbeatIntervalNanos,
+            TimeUnit.NANOSECONDS);
+  }
+
+  private void sendHeartbeatDetails(Object details) throws TException {
```
mfateev: The method name is confusing if there are not details. What about sendHeartbeatRequest?

src/main/java/com/uber/cadence/internal/sync/ActivityExecutionContextImpl.java
```diff
@@ -45,28 +49,88 @@
...
+    this.heartbeatExecutor = heartbeatExecutor;
   }
 
-  /**
-   * @throws CancellationException
-   * @see ActivityExecutionContext#recordActivityHeartbeat(Object)
-   */
+  /** @see ActivityExecutionContext#recordActivityHeartbeat(Object) */
   @Override
   public void recordActivityHeartbeat(Object details) throws ActivityCompletionException {
```
mfateev: details can be null. I think logic is broken in this case.

src/main/java/com/uber/cadence/internal/sync/ActivityExecutionContextImpl.java
```diff
@@ -45,28 +49,88 @@
   private static final Logger log = LoggerFactory.getLogger(ActivityExecutionContextImpl.class);
 
   private final IWorkflowService service;
-
   private final String domain;
-
   private final ActivityTask task;
   private final DataConverter dataConverter;
   private boolean doNotCompleteOnReturn;
+  private final long heartbeatIntervalNanos;
```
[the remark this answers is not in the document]
meiliang86: schedule method in executor service does not take duration directly, so I'll change to use millis.

src/main/java/com/uber/cadence/internal/sync/ActivityExecutionContextImpl.java
meiliang86: Yes, log context has this. This was copied from old code. I'll clean up.

src/main/java/com/uber/cadence/internal/sync/ActivityExecutionContextImpl.java
meiliang86: Done.

src/main/java/com/uber/cadence/internal/sync/ActivityExecutionContextImpl.java
meiliang86: Sure.

src/main/java/com/uber/cadence/internal/sync/ActivityExecutionContextImpl.java
meiliang86: Good catch. I fixed the logic and add a new unit test to cover this logic.

PR #180, opened 2018-06-14
add metric for corrupted signal

halakaraki:


src/test/java/com/uber/cadence/workflow/interceptors/SignalWorkflowInterceptor.java
```diff
@@ -0,0 +1,127 @@
...
+import java.time.Duration;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Random;
+import java.util.UUID;
+import java.util.function.BiPredicate;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+public class SignalWorkflowInterceptor implements WorkflowInterceptor {
```
mfateev: I don't think this class and the factory are generally useful. Let's move them under the test.

src/test/java/com/uber/cadence/workflow/samples/ReceiveSignalObject_ChildWorkflow.java
```diff
@@ -0,0 +1,34 @@
...
+ *  permissions and limitations under the License.
+ */
+
+package com.uber.cadence.workflow.samples;
+
+import com.uber.cadence.workflow.SignalMethod;
+import com.uber.cadence.workflow.WorkflowMethod;
+import com.uber.cadence.workflow.samples.Signal;
+
+public interface ReceiveSignalObject_ChildWorkflow {
```
mfateev: We have separate package for samples. Create PR for it instead of this one. https://github.com/mfateev/uber-java-cadence-samples

halakaraki: renamed to workflows

src/test/java/com/uber/cadence/workflow/interceptors/SignalWorkflowInterceptor.java
halakaraki: The ability to override signals is potentially useful. The factory is test specific, it overrides the signal in a very specific way to the test so I moved it there.

src/test/java/com/uber/cadence/workflow/workflows/SendSignalObject_WorkflowImpl.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package com.uber.cadence.workflow.workflows;
```
meiliang86: Minor comment: One implementation per file might be too verbose. It might make sense to keep a set of related implementations in one file.

meiliang86:
@halakaraki There's one metrics test that is failing and seems to be related to the change. Please fix that before merge. You should be able to see the failing test if you click the travis-ci link.

src/test/java/com/uber/cadence/workflow/workflows/SendSignalObject_WorkflowImpl.java
halakaraki: Do we have some sort of rough guidance on how we organize our classes/files? i'm used to one class per file. Relevant ones live in the same directory then they're collapsable which keeps the the noise in the project view low. At the same time content of the file and available classes are obvious from just expanding and looking at that directory

mfateev: The default is one class per file.  But if some class is used only as an implementation helper for another class then it is usually embedded in that class. 
Unit tests tend to embed all classes that the given unit test needs.

halakaraki: We have multiple tests living in one file. With every test adding it's own helper methods/classes as members of the same test class, the class grows really large. It also discourages visibility and reuse as new tests get written. 
LoggerTest.TestWorkflow, MetricsTest.TestWorkflow1, MetricsTest.TextExceptionPropagation, MetricsTest.ITestChild interfaces look the same.

mfateev: I see no problem having single MetricTest.Workflow interface instead of three. But at the same time there is not that much code duplication even now. I don't think that moving all of these dozens of workflows that are used by a single test into separate files going to improve visibility.

halakaraki: moved the classes into the test file

src/test/java/com/uber/cadence/workflow/MetricsTest.java
```diff
@@ -108,21 +104,79 @@ public void executeChild() {
     }
   }
 
-  @Before
-  public void setUp() {
+  public interface ReceiveSignalObject_ChildWorkflow {
```
mfateev: Java convention is to use CamelCase for Class names. No underscores allowed.http://www.oracle.com/technetwork/java/codeconventions-135099.html

PR #182, opened 2018-06-21
Fixed doNotRetry RetryOptions

mfateev:
Added type adapter for Class objects to DataConverter

### DIFF
--- a/src/main/java/com/uber/cadence/converter/JsonDataConverter.java
+++ b/src/main/java/com/uber/cadence/converter/JsonDataConverter.java
@@ -214,7 +214,7 @@ public final class JsonDataConverter implements DataConverter {
           @Override
           public void write(JsonWriter out, T value) throws IOException {
             out.beginObject();
-            String className = value.getClass().getName();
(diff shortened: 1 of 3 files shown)


src/main/java/com/uber/cadence/converter/JsonDataConverter.java
```diff
@@ -208,6 +209,35 @@ public T read(JsonReader in) throws IOException {
           }
         };
       }
+      if (Class.class.isAssignableFrom(typeToken.getRawType())) {
+        return new TypeAdapter<T>() {
+          @Override
+          public void write(JsonWriter out, T value) throws IOException {
+            out.beginObject();
+            String className = value.getClass().getName();
```
meiliang86: This should be ((Class) value).getName();
value.getClass().getName() is always java.lang.Class
Let's add a DataConverter to test this directly?

mfateev: Good catch!

PR #183, opened 2018-06-26
Changed DataConverterException serialization.

mfateev:
Removed field of type Type from DataConverterException to avoid its serialization.
Fixed handling of empty content in JsonDataConverter.
Fixed handling of empty input in startWorkflow in TestWorkflowService.

### DIFF
--- a/src/main/java/com/uber/cadence/converter/DataConverterException.java
+++ b/src/main/java/com/uber/cadence/converter/DataConverterException.java
@@ -57,11 +57,11 @@ public class DataConverterException extends RuntimeException {
   }
 
   private static String truncateContent(byte[] content) {
-    if (content != null) {
-      // Limit size of the string.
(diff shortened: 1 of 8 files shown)


src/main/java/com/uber/cadence/converter/DataConverterException.java
```diff
@@ -28,42 +28,40 @@
...
 
-  @Override
-  public String getMessage() {
-    if (content == null && valueTypes == null) {
-      return super.getMessage();
+  private static String truncateContent(byte[] content) {
+    if (content != null) {
+      // Limit size of the string.
+      int maxIndex = Math.min(content.length, 255);
+      return new String(content, 0, maxIndex, StandardCharsets.UTF_8);
```
meiliang86: Put "truncated" at the end?

PR #184, opened 2018-06-29
DataConvertor Argument Deserialization: Use defaults for missing arguments and ignore additional arguments

halakaraki:


src/main/java/com/uber/cadence/converter/JsonDataConverter.java
```diff
@@ -165,7 +166,12 @@ public JsonDataConverter(Function<GsonBuilder, GsonBuilder> builderInterceptor)
       JsonArray array = parser.parse(new String(content, StandardCharsets.UTF_8)).getAsJsonArray();
       Object[] result = new Object[valueTypes.length];
       for (int i = 0; i < valueTypes.length; i++) {
-        result[i] = gson.fromJson(array.get(i), valueTypes[i]);
+
+        if(i >= array.size()){ //Missing arugments => add defaults
+          result[i] = Defaults.defaultValue(valueTypes[i].getClass());
```
mfateev: valueTypes[i].getClass is always going to return java.lang.reflect.Type class.
I think you want to cast Type to Class and if not possible use null as a default.

src/test/java/com/uber/cadence/converter/JsonDataConverterTest.java
```diff
@@ -75,6 +75,42 @@ public void testUUIDList() throws NoSuchMethodException {
...
+    Type[] arg = m.getGenericParameterTypes();
+
+    DataConverter converter = JsonDataConverter.getInstance();
+    byte[] data = converter.toData(1,2);
+    @SuppressWarnings("unchecked")
+    Object[] deserializedArguments = converter.fromDataArray(data, arg);
+    assertEquals(5, deserializedArguments.length);
+    assertEquals(1, (int)deserializedArguments[0]);
+    assertEquals(2, (int)deserializedArguments[1]);
+    assertEquals(null, deserializedArguments[2]);
```
mfateev: I would change one of the default arguments to a value type (like int) to see if it is not going to fail.

PR #186, opened 2018-07-06
Thread leaks2

mfateev:
Changed core sizes of executors to 0 to not keep threads around when they are not needed.
Changed executor shutdown to shutdownNow to interrupt waiting threads.
After this change unit tests do not run out of threads.

### DIFF
--- a/src/main/java/com/uber/cadence/internal/sync/DeterministicRunnerImpl.java
+++ b/src/main/java/com/uber/cadence/internal/sync/DeterministicRunnerImpl.java
@@ -128,7 +128,7 @@ class DeterministicRunnerImpl implements DeterministicRunner {
         new ThreadFactory() {
           @Override
           public Thread newThread(Runnable r) {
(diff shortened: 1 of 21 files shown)


src/main/java/com/uber/cadence/internal/sync/DeterministicRunnerImpl.java
```diff
@@ -121,11 +118,20 @@ static void setCurrentThreadInternal(WorkflowThread coroutine) {
...
+  }
+
+  private static ThreadPoolExecutor getDefaultThreadPool() {
+    ThreadPoolExecutor result =
+        new ThreadPoolExecutor(0, 1000, 1, TimeUnit.SECONDS, new SynchronousQueue<>());
+    result.setThreadFactory(
+        new ThreadFactory() {
+          @Override
+          public Thread newThread(Runnable r) {
+            return new Thread(r, "determinstic runner thread");
```
meiliang86: typo

PR #187, opened 2018-07-12
add LruCache

halakaraki:


src/main/java/com/uber/cadence/internal/common/Cache.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package com.uber.cadence.internal.common;
+
+// A Cache is a generalized interface to a cache
+public interface Cache<T> {
+
+  // Get retrieves an element based on a key, returning nil if the element
+  // does not exist
+  T Get(String key);
```
mfateev: Use Java naming conventions. Methods start from lower case letter.
http://www.oracle.com/technetwork/java/javase/overview/codeconventions-135099.html
https://google.github.io/styleguide/javaguide.html

mfateev: I would make key type a generic parameter as well. Look at Map interface.

src/main/java/com/uber/cadence/internal/common/LruCache.java
```diff
@@ -0,0 +1,148 @@
...
+package com.uber.cadence.internal.common;
+
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Consumer;
+
+public class LruCache<T> implements Cache<T> {
```
mfateev: public final ...

src/main/java/com/uber/cadence/internal/common/LruCache.java
```diff
@@ -0,0 +1,148 @@
...
+
+  private int MAX_ENTRIES;
+  private Consumer<T> actOnEvictedEntry = e -> {};
+  private Lock lock = new ReentrantLock();
```
mfateev: private final

src/main/java/com/uber/cadence/internal/common/LruCache.java
```diff
@@ -0,0 +1,148 @@
...
+        protected @Override boolean removeEldestEntry(Map.Entry eldest) {
+          if (size() > MAX_ENTRIES) {
+            @SuppressWarnings("unchecked")
+            T entry = (T) (eldest.getValue());
+            actOnEvictedEntry.accept(entry);
+            return true;
+          }
+          return false;
+        }
+      };
```
mfateev: It should be initialized in the constructor for MAX_ENTRIES to be used.

src/main/java/com/uber/cadence/internal/common/LruCache.java
```diff
@@ -0,0 +1,148 @@
...
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Consumer;
+
+public class LruCache<T> implements Cache<T> {
+
+  private int MAX_ENTRIES;
+  private Consumer<T> actOnEvictedEntry = e -> {};
```
mfateev: Name "e" is usually used by exceptions. This doesn't look like exception handler.

src/main/java/com/uber/cadence/internal/common/LruCache.java
```diff
@@ -0,0 +1,148 @@
...
+    lock.lock();
+    try {
+      map.put(key, value);
+    } finally {
+      lock.unlock();
+    }
+  }
+
+  @Override
+  public void PutIfNotExist(String key, T value) {
```
mfateev: Note that you use similar, but different name than the one already used by the Map. I would just stick to the Map name (putIfAbsent).

src/main/java/com/uber/cadence/internal/common/LruCache.java
```diff
@@ -0,0 +1,148 @@
...
+  public int Size() {
+    lock.lock();
+    try {
+      return map.size();
+    } finally {
+      lock.unlock();
+    }
+  }
+
+  private void deleteNoLock(String key) {
```
mfateev: Rename to deleteLocked

src/main/java/com/uber/cadence/internal/common/LruCache.java
```diff
@@ -0,0 +1,148 @@
...
+    T value = map.remove(key);
+    actOnEvictedEntry.accept(value);
```
mfateev: Is it OK to pass null into accept here? What's the behaviour of delete for non existent key?

src/main/java/com/uber/cadence/internal/replay/ReplayDecider.java
```diff
@@ -401,4 +401,8 @@ DecisionsHelper getDecisionsHelper() {
     decideImpl(() -> result.set(workflow.query(query)));
     return result.get();
   }
+
+  public void Close() {
```
mfateev: Invalid name.

src/main/java/com/uber/cadence/internal/replay/ReplayDeciderCache.java
```diff
@@ -0,0 +1,27 @@
...
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package com.uber.cadence.internal.replay;
+
+import com.uber.cadence.internal.common.LruCache;
+
+class ReplayDeciderCache extends LruCache<ReplayDecider> {
```
mfateev: Don't use inheritance to configure. Just instantiate LruCache with appropriate constructor parameters.

src/test/java/com/uber/cadence/internal/common/LruCacheTests.java
```diff
@@ -0,0 +1,119 @@
...
+    c.Put("key3", "c");
+    c.Get("key1");
+    c.Put("key4", "c");
+
+    String shouldBeEvicted = c.Get("key2");
+    String shouldNotBeEvicted = c.Get("key1");
+
+    assertEquals(3, c.Size());
+    assertEquals("a", shouldNotBeEvicted);
+    assertEquals(null, shouldBeEvicted);
```
mfateev: Use assertNull

src/test/java/com/uber/cadence/internal/common/LruCacheTests.java
```diff
@@ -0,0 +1,119 @@
...
+
+    // Verify actOnEvict is called on eviction
+    assertEquals(1, elementsEvicted.get());
+  }
+
+  @Test
+  public void LeastRecentlyUsedEntryIsEvictedOnEvictNext() {
+    AtomicInteger elementsEvicted = new AtomicInteger(0);
```
mfateev: Another option is to use a set of evicted and assert against it.

meiliang86:
Do I understand correctly that most of the changes in this PR is not needed if we use some off the shelf LRU cache implementation?

PR #191, opened 2018-08-03
Refactor replay decider - Remove dependency on requiring full history for decider.Decide()

halakaraki:


src/main/java/com/uber/cadence/internal/worker/WorkflowWorker.java
```diff
@@ -250,21 +249,17 @@ private void sendReply(
 
   private class DecisionTaskWithHistoryIteratorImpl implements DecisionTaskWithHistoryIterator {
 
-    private long start = System.currentTimeMillis();
+    private final Duration RetryServiceOperationInitialInterval = Duration.ofMillis(200);
+    private final Duration RetryServiceOperationMaxInterval = Duration.ofSeconds(4);
+    private final Duration RetryServiceOperationExpirationInterval = Duration.ofSeconds(60);
```
mfateev: I think it should be equal to the decision task timeout.

src/main/java/com/uber/cadence/internal/worker/WorkflowWorker.java
```diff
@@ -287,21 +282,15 @@ public HistoryEvent next() {
...
 
           options.getMetricsScope().counter(MetricsType.WORKFLOW_GET_HISTORY_COUNTER).inc(1);
           Stopwatch sw =
               options.getMetricsScope().timer(MetricsType.WORKFLOW_GET_HISTORY_LATENCY).start();
           RetryOptions retryOptions =
               new RetryOptions.Builder()
-                  .setExpiration(expiration)
-                  .setInitialInterval(Duration.ofMillis(50))
-                  .setMaximumInterval(Duration.ofSeconds(1))
+                  .setExpiration(RetryServiceOperationExpirationInterval)
```
mfateev: Why did you remote the expiration calculation from the decision task timeout? Retrying for a minute when decision task has 10 second timeout is incorrect.

halakaraki: changed it to use the timeout on the DecisionTaskScheduled Event

src/main/java/com/uber/cadence/internal/replay/ReplayDecider.java
```diff
@@ -17,12 +17,7 @@
 
 package com.uber.cadence.internal.replay;
 
-import com.uber.cadence.EventType;
-import com.uber.cadence.HistoryEvent;
-import com.uber.cadence.PollForDecisionTaskResponse;
-import com.uber.cadence.TimerFiredEventAttributes;
-import com.uber.cadence.WorkflowExecutionSignaledEventAttributes;
-import com.uber.cadence.WorkflowQuery;
+import com.uber.cadence.*;
```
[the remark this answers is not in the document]
halakaraki: No, i used WorkflowExecutionStartedEventAttributes and the ide decided to import cadence.* instead of just that class. Fixed it

src/main/java/com/uber/cadence/internal/worker/WorkflowWorker.java
```diff
@@ -250,23 +240,35 @@ private void sendReply(
 
   private class DecisionTaskWithHistoryIteratorImpl implements DecisionTaskWithHistoryIterator {
 
-    private long start = System.currentTimeMillis();
+    private final Duration RetryServiceOperationInitialInterval = Duration.ofMillis(200);
```
mfateev: Field and method names in Java are always start from lower case letter. The exception is CONSTANTS_THAT_USE_THIS_STYLE

src/main/java/com/uber/cadence/internal/worker/WorkflowWorker.java
```diff
@@ -250,23 +240,35 @@ private void sendReply(
...
+    private final Duration RetryServiceOperationMaxInterval = Duration.ofSeconds(4);
+    private final Duration PaginationStart = Duration.ofMillis(System.currentTimeMillis());
+    private Duration DecisionTaskStartToCloseTimeout = Duration.ofSeconds(60);
```
mfateev: Where is this coming from? There is no default decision task timeout. It should come from the decision task.

src/main/java/com/uber/cadence/internal/worker/WorkflowWorker.java
halakaraki: Resovled

src/main/java/com/uber/cadence/internal/worker/WorkflowWorker.java
halakaraki: removed default value. This method will now throw if the response does not contain decision task scheduled

src/main/java/com/uber/cadence/internal/worker/WorkflowWorker.java
```diff
@@ -240,14 +240,14 @@ private void sendReply(
...
   private class DecisionTaskWithHistoryIteratorImpl implements DecisionTaskWithHistoryIterator {
 
-    private final Duration RetryServiceOperationInitialInterval = Duration.ofMillis(200);
-    private final Duration RetryServiceOperationMaxInterval = Duration.ofSeconds(4);
-    private final Duration PaginationStart = Duration.ofMillis(System.currentTimeMillis());
-    private Duration DecisionTaskStartToCloseTimeout = Duration.ofSeconds(60);
+    private final Duration retryServiceOperationInitialInterval = Duration.ofMillis(200);
+    private final Duration retryServiceOperationMaxInterval = Duration.ofSeconds(4);
+    private final Duration paginationStart = Duration.ofMillis(System.currentTimeMillis());
+    private Duration decisionTaskStartToCloseTimeout = null;
```
mfateev: No need to initialize to the default value.

src/main/java/com/uber/cadence/internal/worker/WorkflowWorker.java
```diff
@@ -240,14 +240,14 @@ private void sendReply(
...
 
     private final Duration RetryServiceOperationExpirationInterval() {
```
mfateev: Method name should start from a lower case letter.

halakaraki: resolved

PR #194, opened 2018-08-07
fix ConcurrentModificationException in DeterministicRunnerImpl

halakaraki:


src/main/java/com/uber/cadence/internal/sync/DeterministicRunnerImpl.java
```diff
@@ -282,7 +282,8 @@ public void close() {
         c.stop();
       }
       threads.clear();
-      for (Promise<?> f : failedPromises) {
+      Set<Promise> failedPromisesLoop = new HashSet<>(failedPromises);
```
meiliang86: Can you add a comment here to explain why we did the copy? Thanks!

halakaraki: Will do

PR #196, opened 2018-08-21
Fix test environment

halakaraki:
Add Support for Sticky Execution Attributes
FIx event ids so that they start from 1


src/test/java/com/uber/cadence/testUtils/TestServiceUtils.java
```diff
@@ -0,0 +1,100 @@
...
+    request.taskList = createNormalTaskList(tasklistName);
+    request.setExecutionStartToCloseTimeoutSeconds(executionStartToCloseTimeoutSeconds);
+    request.setTaskStartToCloseTimeoutSeconds(taskStartToCloseTimeoutSeconds);
+    WorkflowType type = new WorkflowType();
+    type.name = workflowType;
+    request.workflowType = type;
+    service.StartWorkflowExecution(request);
+    }
+
+    public static void respondDecisionTaskCompletedWithSticky(
```
meiliang86: This overload does not seem to be very useful. IMO it's better to just keep the other version and use it.

src/main/java/com/uber/cadence/internal/testservice/TestWorkflowService.java
```diff
@@ -266,13 +266,15 @@ public PollForDecisionTaskResponse PollForDecisionTask(PollForDecisionTaskReques
     TestWorkflowMutableState mutableState = getMutableState(executionId);
     try {
       mutableState.startDecisionTask(task, pollRequest);
+      task.setWorkflowExecutionTaskList(mutableState.getStartRequest().taskList);
```
meiliang86: This can be different than the task list that is used actually for pulling.
For example check taskCompletionWithStickyExecutionAttributesWillScheduleDecisionsOnStickyTaskList.

src/test/java/com/uber/cadence/testUtils/TestServiceUtils.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.ArrayList;
+import java.util.UUID;
+
+import static com.uber.cadence.internal.common.InternalUtils.createNormalTaskList;
+import static com.uber.cadence.internal.common.InternalUtils.createStickyTaskList;
+
+public class TestServiceUtils{
+    private TestServiceUtils() {}
+
+    public static void startWorkflowExecution(
```
meiliang86: Same here. Let's just keep the other version and leave the constants in the other test file.

src/test/java/com/uber/cadence/internal/testing/WorkflowStickynessTest.java
```diff
@@ -0,0 +1,107 @@
...
+public class WorkflowStickynessTest {
+
+  @Test
+  public void taskCompletionWithStickyExecutionAttributesWillScheduleDecisionsOnStickyTaskList()
+      throws Exception {
+
+    TestWorkflowService service = new TestWorkflowService();
+    service.lockTimeSkipping("HistoryUtils");
+
+    TestServiceUtils.startWorkflowExecution("domain", "tasklist", "wfType", service);
```
meiliang86: Define some constants for these literals?

src/test/java/com/uber/cadence/internal/testing/WorkflowStickynessTest.java
```diff
@@ -0,0 +1,107 @@
...
+        response.taskToken, "tasklist-sticky", service);
+    TestServiceUtils.signalWorkflow(response.workflowExecution, "domain", service);
+    response =
+        TestServiceUtils.pollForDecisionTask(
+            "domain", createStickyTaskList("tasklist-sticky"), service);
+    TestServiceUtils.respondDecisionTaskFailedWithSticky(response.taskToken, service);
+    response =
+        TestServiceUtils.pollForDecisionTask("domain", createStickyTaskList("tasklist"), service);
+
+    assertEquals(10, response.history.getEventsSize());
```
meiliang86: This number 10 looks like a magic number to me. Maybe we just check list is non-empty and details in the last event?

src/test/java/com/uber/cadence/testUtils/TestServiceUtils.java
halakaraki: resolved.

src/main/java/com/uber/cadence/internal/testservice/TestWorkflowService.java
halakaraki: added a comment

src/test/java/com/uber/cadence/internal/testing/WorkflowStickynessTest.java
halakaraki: resolved

PR #201, opened 2018-08-31
1. Refactor decider

halakaraki:


src/main/java/com/uber/cadence/internal/replay/Decider.java
```diff
@@ -17,13 +17,14 @@
...
+import com.uber.cadence.Decision;
 import com.uber.cadence.WorkflowQuery;
 import com.uber.cadence.internal.worker.DecisionTaskWithHistoryIterator;
+import java.util.List;
 
 public interface Decider {
 
-  // TODO: refactor in future CR. Merge methods and decide should return a list of decisions.
-  void decide(DecisionTaskWithHistoryIterator iterator) throws Throwable;
+  List<Decision> decide(DecisionTaskWithHistoryIterator iterator) throws Throwable;
```
halakaraki: I kept query and decide as separate methods since they yield very different results that the handler has to deal with differently.
The ideal situation is that each method only takes the exact input it needs instead of the whole iterator. This requires further thought/might be a bigger refactor.

PR #202, opened 2018-09-01
2. Wire sticky poller and cache

halakaraki:


src/test/java/com/uber/cadence/worker/StickyWorkerTest.java
```diff
@@ -0,0 +1,325 @@
...
+    assertEquals(workflow.getProgress(), GreetingWorkflow.Status.WAITING_FOR_NAME);
+
+    workflow.waitForName("World");
+    String greeting = workflow.getGreeting();
+
+    assertEquals("Hello World!", greeting);
+    assertEquals(workflow.getProgress(), GreetingWorkflow.Status.GREETING_GENERATED);
+    wrapper.close();
+  }
+
```
mfateev: Let's add the test for the sticky task schedule to start timeout.

src/main/java/com/uber/cadence/internal/worker/WorkflowWorker.java
```diff
@@ -243,7 +263,7 @@ private void sendReply(
     private final Duration retryServiceOperationInitialInterval = Duration.ofMillis(200);
     private final Duration retryServiceOperationMaxInterval = Duration.ofSeconds(4);
     private final Duration paginationStart = Duration.ofMillis(System.currentTimeMillis());
-    private Duration decisionTaskStartToCloseTimeout;
+    private Duration decisionTaskStartToCloseTimeout = Duration.ofSeconds(60);
```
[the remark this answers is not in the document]
halakaraki: Removed the initialization.
Currently DecisionTaskWithHistoryIteratorImpl does not have access to WorkflowExecutionStarted event at construction time. Replay decider will have to pass it in.
Options:
1. Refactor the interface DecisionTaskHandler to not take a dependency on HistoryIterator and just consume DecisionTask instead. Because of how queries work and ReplayDecisionTaskWithHistoryIterator work this is a bit more involved refactor.

2. Pass WorkflowExecutionStarted at getHistory(). However, the requirement to pass this in is implementation specific and it doesn't seem appropriate to enforce it at the interface.

3. provide a separate method to pass it. This feels like a bad pattern as now it requires the user of this class to know that they need to call this method prior to calling getHistory();

4. Current solution. Abstracted it out in a method and only call that method in get history if we need to paginate. Best option IMO is 1. but I would rather do that in a separate CR

src/main/java/com/uber/cadence/worker/Worker.java
```diff
@@ -21,27 +21,31 @@
...
+import com.uber.cadence.PollForDecisionTaskResponse;
 import com.uber.cadence.WorkflowExecution;
 import com.uber.cadence.client.WorkflowClient;
 import com.uber.cadence.converter.DataConverter;
 import com.uber.cadence.internal.metrics.MetricsTag;
+import com.uber.cadence.internal.replay.DeciderCache;
 import com.uber.cadence.internal.sync.SyncActivityWorker;
 import com.uber.cadence.internal.sync.SyncWorkflowWorker;
-import com.uber.cadence.internal.worker.SingleWorkerOptions;
+import com.uber.cadence.internal.worker.*;
```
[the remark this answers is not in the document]
halakaraki: runs with commandline build but it doesn't remove these. I have some setting in intellij that does that when it automatically imports libraries. I'll figure out where that is and resolve it.

src/main/java/com/uber/cadence/worker/Worker.java
```diff
@@ -305,30 +328,87 @@ public String getTaskList() {
   }
 
   public static final class Factory {
-
     private final List<Worker> workers = new ArrayList<>();
     private final IWorkflowService workflowService;
     private final String domain;
+    private final UUID Id = UUID.randomUUID();
```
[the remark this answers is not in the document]
halakaraki: It's an id to guarantee uniqueness incase multiple factories are created in the same host to not share a task list. Not sure what else to name it. I could add a comment next to it

src/main/java/com/uber/cadence/worker/Worker.java
```diff
@@ -305,30 +328,87 @@ public String getTaskList() {
...
+    private final ThreadPoolExecutor workflowThreadPool;
+    private final AtomicInteger workflowThreadCounter = new AtomicInteger();
+
+    private FactoryOptions factoryOptions;
+    private Poller<PollForDecisionTaskResponse> stickyPoller;
+    private Dispatcher<String, PollForDecisionTaskResponse> dispatcher;
+    private DeciderCache cache;
+
```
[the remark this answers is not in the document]
halakaraki: Poller, Dispatcher, and Cache are only instantiated if stickyness is enabled. So can't make them final even though they're only instantiated in the constructor

src/test/java/com/uber/cadence/worker/StickyWorkerTest.java
halakaraki: There is one already. Take a look at ReplayDeciderTaskHandlerTests.ifStickyExecutionAttributesAreSetThenWorkflowsAreCached()

src/main/java/com/uber/cadence/internal/worker/WorkflowPollTaskFactory.java
```diff
@@ -0,0 +1,46 @@
...
+
+package com.uber.cadence.internal.worker;
+
+import com.uber.cadence.PollForDecisionTaskResponse;
+import com.uber.cadence.serviceclient.IWorkflowService;
+import java.util.Objects;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class WorkflowPollTaskFactory implements PollTaskFactory<PollForDecisionTaskResponse> {
```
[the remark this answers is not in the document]
halakaraki: I'm confused, as in the Factory can be replaced with a lambda? Wouldn't that necessitate WorkflowPollTask being public? WorkflowPollTask

src/main/java/com/uber/cadence/worker/Worker.java
```diff
@@ -64,13 +71,26 @@
...
    */
-  private Worker(IWorkflowService service, String domain, String taskList, WorkerOptions options) {
+  private Worker(
+      IWorkflowService service,
+      String domain,
+      String taskList,
+      WorkerOptions options,
+      DeciderCache cache,
+      String stickyTaskListName,
+      Duration stickyDecisionScheduleToStartTimeout,
```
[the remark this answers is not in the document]
halakaraki: yes, but for the most part it's expected that worker wouldn't need them since these options will be to instantiate/set objects that are outside the worker's scope (cache and poller). I can refactor then if there's a pattern of having to pass them in.

src/main/java/com/uber/cadence/internal/common/WorkflowExecutionUtils.java
```diff
@@ -553,60 +553,6 @@ public static WorkflowExecutionInfo describeWorkflowInstance(
     return instanceMetadata;
   }
 
-  /** Returns workflow instance history in a human readable format. */
```
mfateev: Why is it removed?

src/main/java/com/uber/cadence/internal/replay/ReplayDecider.java
```diff
@@ -419,10 +436,86 @@ DecisionsHelper getDecisionsHelper() {
...
-  public byte[] query(DecisionTaskWithHistoryIterator decisionTaskIterator, WorkflowQuery query)
-      throws Throwable {
+  public byte[] query(PollForDecisionTaskResponse response, WorkflowQuery query) throws Throwable {
     AtomicReference<byte[]> result = new AtomicReference<>();
-    decideImpl(decisionTaskIterator, () -> result.set(workflow.query(query)));
+    decideImpl(response, () -> result.set(workflow.query(query)));
     return result.get();
   }
+
+  class DecisionTaskWithHistoryIteratorImpl implements DecisionTaskWithHistoryIterator {
```
mfateev: Could you make it private?

src/main/java/com/uber/cadence/worker/Worker.java
```diff
@@ -367,23 +473,116 @@ public void start() {
...
+        return new FactoryOptions(
+            enableStickyExecution,
+            cacheMaximumSize,
+            maxWorkflowThreadCount,
+            stickyDecisionScheduleToStartTimeoutInSeconds);
+      }
+    }
+
+    private final boolean enableStickyExecution;
+    private int cacheMaximumSize;
```
mfateev: All of them should be final

src/main/java/com/uber/cadence/worker/Worker.java
```diff
@@ -367,23 +473,116 @@ public void start() {
...
+    private int maxWorkflowThreadCount;
+    private final int stickyDecisionScheduleToStartTimeoutInSeconds;
+
+    private FactoryOptions(
+        boolean enableStickyExecution,
+        int cacheMaximumSize,
+        int maxWorkflowThreadCount,
+        int stickyDecisionScheduleToStartTimeoutInSeconds) {
+      this.enableStickyExecution = enableStickyExecution;
+      this.cacheMaximumSize = cacheMaximumSize;
```
mfateev: Add checks for valid values.

src/main/java/com/uber/cadence/worker/Worker.java

src/main/java/com/uber/cadence/internal/replay/ReplayDecider.java
halakaraki: resolved

src/main/java/com/uber/cadence/internal/common/WorkflowExecutionUtils.java
halakaraki: Because it's unused and in an internal library. There's another overload which prints the history in a pretty format given the history. Do we need a print method that fetches the history on it's own? I can add it back but i'm not sure i see the purpose.

mfateev: Let's keep it as it can be used by tools. For example https://github.com/uber/cadence-java-samples/blob/master/src/main/java/com/uber/cadence/samples/common/WorkflowExecutionHistoryPrinter.java

halakaraki: ok, reverted

PR #204, opened 2018-09-12
Sticky metrics

halakaraki:


src/main/java/com/uber/cadence/worker/Worker.java
```diff
@@ -558,25 +565,41 @@ public Builder setStickyDecisionScheduleToStartTimeoutInSeconds(
...
+            stickyWorkflowPollerOptions,
+            metricScope);
       }
     }
 
     private final boolean enableStickyExecution;
     private final int cacheMaximumSize;
     private final int maxWorkflowThreadCount;
     private final int stickyDecisionScheduleToStartTimeoutInSeconds;
+    private PollerOptions stickyWorkflowPollerOptions;
```
mfateev: all fields should be final

src/test/java/com/uber/cadence/internal/replay/ReplayDeciderCacheTests.java
```diff
@@ -17,27 +17,36 @@
...
 import com.uber.cadence.worker.WorkerOptions;
+import com.uber.m3.tally.RootScopeBuilder;
+import com.uber.m3.tally.Scope;
+import com.uber.m3.tally.StatsReporter;
+import com.uber.m3.util.Duration;
+import com.uber.m3.util.ImmutableMap;
 import junit.framework.TestCase;
 import org.junit.Test;
 
+import java.util.Map;
```
mfateev: stars and spaces

src/main/java/com/uber/cadence/internal/metrics/MetricsType.java
```diff
@@ -134,5 +134,8 @@
 
   public static final String STICKY_CACHE_HIT = CADENCE_METRICS_PREFIX + "sticky-cache-hit";
   public static final String STICKY_CACHE_MISS = CADENCE_METRICS_PREFIX + "sticky-cache-miss";
+  public static final String STICKY_CACHE_FORCED_EVICTION =
+      CADENCE_METRICS_PREFIX + "sticky-cache-forced-eviction";
```
mfateev: I think we want two metrics. One for thread forced another for count forced. May be one total another thread only.

halakaraki: added one specifically for threads and the existing one will count the total

src/main/java/com/uber/cadence/worker/Worker.java

src/test/java/com/uber/cadence/internal/replay/ReplayDeciderCacheTests.java
halakaraki: resolved

PR #210, opened 2018-09-25
Allow passing additional TChannel transport headers

beaverlyhills:


src/main/java/com/uber/cadence/serviceclient/WorkflowServiceTChannel.java
```diff
@@ -131,6 +131,9 @@
     /** Client for metrics reporting. */
     private final Scope metricsScope;
 
+    /** Optional TChannel transport headers */
+    private final Map<String, String> transportHeaders;
```
halakaraki: What's the scenario you're adding the custom headers for?

PR #211, opened 2018-09-27
Reset Replay Decider State

halakaraki:


src/main/java/com/uber/cadence/internal/replay/HistoryHelper.java
```diff
@@ -196,12 +200,17 @@ public DecisionEvents next() {
...
       boolean replay = true;
-      long replayCurrentTimeMilliseconds = -1;
       long nextDecisionEventId = -1;
       while (events.hasNext()) {
         HistoryEvent event = events.next();
         EventType eventType = event.getEventType();
-        // hasNext is for queries that do not have DecisionTaskStarted at the end of the history.
+
+        // Sticky workers receive an event history that starts with DecisionTaskCompleted
+        if (eventType == EventType.DecisionTaskCompleted) {
```
mfateev: Shouldn't the condition be 
`eventType == EventType.DecisionTaskCompleted && nextDecisionEventId == -1`
?

src/main/java/com/uber/cadence/internal/replay/HistoryHelper.java
```diff
@@ -216,10 +225,14 @@ public DecisionEvents next() {
             continue;
           } else if (peekedType == EventType.DecisionTaskCompleted) {
             events.next(); // consume DecisionTaskCompleted
-            nextDecisionEventId = peeked.getEventId() + 1; // +1 for next
+            nextDecisionEventId = peeked.getEventId() + 1; // +1 for next and skip over completed
```
mfateev: What about checking id of every event?
Something like
```
if (nextDecisionEventId++ != peeked.getEventId()) {
   throw...
}
```

src/main/java/com/uber/cadence/internal/replay/HistoryHelper.java
halakaraki: Yes it should. Well, later completed events always follow started events so it gets handled in the code below. But it's safer/cleaner to add that condition so i'll add it.

src/main/java/com/uber/cadence/internal/replay/HistoryHelper.java
halakaraki: nextDecisionEventId is set at this point. It has no previous value. If we want to validate eventIds being in order I think we should do it in the piece of code that handles pagination (I'll add it to that CR).

PR #212, opened 2018-10-05
set nextPageToken on GetWorkflowExecutionHistoryRequest

halakaraki:


src/main/java/com/uber/cadence/internal/replay/Decider.java
```diff
@@ -29,4 +29,4 @@
   byte[] query(PollForDecisionTaskResponse decisionTask, WorkflowQuery query) throws Throwable;
 
   void close();
-}
+}
```
mfateev: Add new line back

src/test/java/com/uber/cadence/worker/WorkerStressTests.java
```diff
@@ -0,0 +1,201 @@
...
+    @ActivityMethod()
+    void sleep(int chain, int concurrency, byte[] bytes);
+  }
+
+  public static class ActivitiesImpl implements SleepActivity {
+    private static final Logger log = LoggerFactory.getLogger("sleep-activity");
+
+    @Override
+    public void sleep(int chain, int concurrency, byte[] bytes) {
+      log.info("sleep called");
```
mfateev: I'm confused about purpose and name of this activity as it never sleeps.

src/test/java/com/uber/cadence/worker/WorkerStressTests.java
```diff
@@ -0,0 +1,201 @@
...
+                  .setStartToCloseTimeout(Duration.ofMinutes(1))
+                  .setHeartbeatTimeout(Duration.ofSeconds(20))
+                  .build());
+
+      for (int i = 0; i < params.ChainSequence; i++) {
+        List<Promise<Void>> promises = new ArrayList<>();
+        for (int j = 0; j < params.ConcurrentCount; j++) {
+          byte[] bytes = new byte[params.PayloadSizeBytes];
+          new Random().nextBytes(bytes);
+          Promise<Void> promise = Async.procedure(activity::sleep, i, j, bytes);
```
mfateev: This call doesn't create multiple threads as activity implementation is asynchronous in this case.

src/test/java/com/uber/cadence/worker/WorkerStressTests.java
```diff
@@ -0,0 +1,201 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TestName;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@RunWith(Parameterized.class)
+public class WorkerStressTests {
```
mfateev: I think the best way to unit test pagination is to add pagination to the in memory service and set the page size to some low value like 10 events for this test.

src/main/java/com/uber/cadence/internal/replay/Decider.java
halakaraki: added

src/test/java/com/uber/cadence/worker/WorkerStressTests.java
halakaraki: It does on workflow execution in syncDecisionContext. executeActivity will do it in a new workflowThreadImpl so this will cause the threadpool to be consumed

src/test/java/com/uber/cadence/worker/WorkerStressTests.java
halakaraki: renamed it to someActivity. It's purpose is to simply to be triggered, consume threads and generate events

src/test/java/com/uber/cadence/worker/WorkerStressTests.java
mfateev: I'm confused. Are you saying that Async.procedure for activity does consume a thread after all your changes? If it is the case it should be fixed asap.

halakaraki: Discussed offline

src/test/java/com/uber/cadence/worker/WorkerStressTests.java
halakaraki: I'll add a task for this

PR #213, opened 2018-10-08
DeterministicRunnerImpl::close: wait for tasks outside lock

halakaraki:


src/main/java/com/uber/cadence/internal/sync/DeterministicRunnerImpl.java
```diff
@@ -299,7 +300,10 @@ public void close() {
     }
     try {
       for (WorkflowThread c : threads) {
-        c.stop();
+        Future<?> future = c.stopNow();
+        if (future != null) {
```
mfateev: I would make this method always return a future. If it is already stopped it can return a future that is already ready.

halakaraki: Done

PR #214, opened 2018-10-09
Fix evict next

halakaraki:


src/main/java/com/uber/cadence/internal/sync/WorkflowThreadImpl.java
```diff
@@ -238,7 +238,13 @@ public void start() {
           .getMetricsScope()
           .counter(MetricsType.STICKY_CACHE_THREAD_FORCED_EVICTION)
           .inc(1);
-      cache.evictNext();
+      try {
+        if (cache != null) {
+          cache.evictNext();
+        }
+      } catch (InterruptedException e1) {
+        log.warn("Unable to evict cache", e1);
```
mfateev: Is it safe to continue in this case?

halakaraki: It will just attempt to submit again and then fail. Should I throw a specific error here?

PR #219, opened 2018-10-15
Handle Self Eviction

halakaraki:
Block Self Eviction
Evict Next is renamed to EvictAny (Will evict randomly)
Fix warnings that are treated as errors


src/main/java/com/uber/cadence/worker/Worker.java
```diff
@@ -480,6 +544,11 @@ public void start() {
       }
     }
 
+    /**
+     * Shuts down Poller used for sticky workflows and all workers created by this factory.
+     *
+     * @param timeout
```
mfateev: Document what happens on timeout. Does it return or throw?

src/main/java/com/uber/cadence/worker/Worker.java
```diff
@@ -528,28 +597,49 @@ private String getStickyTaskListName() {
       private PollerOptions stickyWorkflowPollerOptions;
       private Scope metricScope;
 
+      /**
+       * When set to true it will create an affinity between the worker and the workflow run it's
+       * processing. Workers will cache workflows and will handle all decisions for that workflow
+       * instance until it's complete or evicted from the cache.
+       */
       public Builder setEnableStickyExecution(boolean enableStickyExecution) {
```
mfateev: Let's change it to setDisableStickyExecution with enabled by default

src/test/java/com/uber/cadence/internal/replay/ReplayDeciderCacheTests.java
```diff
@@ -152,7 +152,7 @@ public void whenHistoryIsPartialAndCacheIsEmptyThenCacheEvictedExceptionIsThrown
   }
 
   @Test
-  public void evictNextWillInvalidateTheNextEntryInLineToBeEvicted() throws Exception {
+  public void evictAnyWillInvalidateAnEntryRandomlyFromTheCache() throws Exception {
```
mfateev: Why  randomly?  Is the cache not a LRU cache?

halakaraki: Evict any is used for thread based eviction. In the normal mode of operation, it is lru. I.e. when elements get kicked out because of exceeding size or weight limit. However if you want to manually evict the next element, the loading cache doesn't support this. The way it was done before was by adding an element large enough to kick the next one in line out. However in this manner you have no control or visibility on which element is evicted and cannot guarantee self eviction doesn't happen.
Options to ensure this that i see is we need to keep track of insertion order but that seems like 
a duplication of effort of what the cache itself is doing (Or write our own cache). I'm not sure though ensuring this will matter on performance but i can file a task

src/main/java/com/uber/cadence/worker/Worker.java

src/main/java/com/uber/cadence/internal/worker/PollDecisionTaskDispatcherFactory.java
```diff
@@ -24,7 +24,7 @@
 
 public final class PollDecisionTaskDispatcherFactory
     implements DispatcherFactory<String, PollForDecisionTaskResponse> {
-  private IWorkflowService service = new WorkflowServiceTChannel();
+  private IWorkflowService service;;
```
mfateev: remove semicolon

src/main/java/com/uber/cadence/worker/Worker.java
```diff
@@ -545,7 +545,9 @@ public void start() {
     }
 
     /**
-     * Shuts down Poller used for sticky workflows and all workers created by this factory.
+     * Shuts down Poller used for sticky workflows and all workers created by this factory. Shutdown
+     * will be called on each worker created by this factory with the given timeout. If the timeout
+     * is exceed the method will thrown an InterruptedException.
```
mfateev: throwing Interrupted on timeout is not a standard Java practice. In case of shutdown returning without error (may be logging a warning) is a preferred option

src/main/java/com/uber/cadence/worker/Worker.java
```diff
@@ -590,20 +593,20 @@ private String getStickyTaskListName() {
 
   public static class FactoryOptions {
     public static class Builder {
-      private boolean enableStickyExecution;
+      private boolean disableStickyExecution = false;
```
mfateev: No need to initialize to a default value.

src/main/java/com/uber/cadence/worker/Worker.java
halakaraki: resovled

PR #222, opened 2018-10-26
Fixed query thread leak

mfateev:


src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -1848,67 +1863,63 @@ public void mySignal(String value) {
 
   @Test
   public void testSignal() {
```
halakaraki: Might be worth splitting this test? I feel it's doing too much it's no longer clear what it's validating.
Maybe a test for: queries don't cause thread leaks (for both sticky and non sticky)
Another test for testing signals? Is the intention here to test signaling or test worker querying? This seems to be doing both and might be worth splitting that too.

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -1848,67 +1863,63 @@ public void mySignal(String value) {
...
+        workflowClient.newWorkflowStub(
+            QueryableWorkflow.class, execution.getWorkflowId(), Optional.of(execution.getRunId()));
+    assertEquals("Hello ", client2.getState());
+
+    String queryResult = null;
+    try {
+      queryResult =
+          queryWorker.queryWorkflowExecution(
+              execution, "QueryableWorkflow::getState", String.class);
+    } catch (Exception e) {
```
halakaraki: Why do this instead of just allowing the test method signature to throw?

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -1848,67 +1863,63 @@ public void mySignal(String value) {
...
+      throw new RuntimeException(e);
+    }
+    assertEquals("Hello ", queryResult);
+    sleep(Duration.ofMillis(500));
+    client2.mySignal("World!");
```
halakaraki: Why are we using 2 workflow clients?

halakaraki:
Can you correct line 231 in WorkflowThreadImpl.java to MetricsType.WORKFLOW_ACTIVE_THREAD_COUNT in this CR?

src/main/java/com/uber/cadence/internal/replay/ReplayDecisionTaskHandler.java
```diff
@@ -152,7 +153,7 @@ private Result processDecision(PollForDecisionTaskResponse decisionTask) throws
       }
```
halakaraki: UI won't let me comment on line 150 but should this be "Throwable"?

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
mfateev: Just to ensure that a client created from workflow id also works correctly.

src/main/java/com/uber/cadence/internal/replay/ReplayDecisionTaskHandler.java
mfateev: Fixed

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
mfateev: Fixed

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
mfateev: Fixed

PR #228, opened 2018-11-10
Added support for server side retry of workflows

mfateev:
- This change adds support for server side workflow retries for start workflow execution only.
- The support for server side retry of activities and child workflows to be added in the following changes.


src/test/java/com/uber/cadence/activity/ActivityOptionsTest.java
```diff
@@ -150,7 +147,6 @@ public void testBothPresent() throws NoSuchMethodException {
                     .setMaximumInterval(Duration.ofDays(3))
                     .setExpiration(Duration.ofDays(365))
                     .setInitialInterval(Duration.ofMinutes(12))
-                    .setMinimumAttempts(2)
                     .build())
             .build();
     ActivityMethod a =
```
meiliang86: I'm not able to leave comments below this line, but should we also get MethodRetry and pass it into merge in the final assertion? i.e.
```
Method method = ActivityOptionsTest.class.getMethod("activityAndRetryOptions");
ActivityMethod a = method.getAnnotation(ActivityMethod.class);
MethodRetry r = method.getAnnotation(MethodRetry.class);
Assert.assertEquals(o, ActivityOptions.merge(a, r, o));
```

src/test/java/com/uber/cadence/activity/ActivityOptionsTest.java
```diff
@@ -57,7 +56,7 @@ public void testOnlyOptionsPresent() throws NoSuchMethodException {
     Assert.assertEquals(o, ActivityOptions.merge(a, null, o));
   }
 
-  @MethodRetry(initialIntervalSeconds = 3)
+  @MethodRetry(initialIntervalSeconds = 3, expirationSeconds = 100)
```
meiliang86: What's the difference between this and the test above? We are not using this annotation anyway.

src/test/java/com/uber/cadence/client/WorkflowOptionsTest.java
```diff
@@ -98,6 +98,6 @@ public void testBothPresent() throws NoSuchMethodException {
             .build();
     WorkflowMethod a =
         WorkflowOptionsTest.class.getMethod("workflowOptions").getAnnotation(WorkflowMethod.class);
-    Assert.assertEquals(o, WorkflowOptions.merge(a, o));
+    Assert.assertEquals(o, WorkflowOptions.merge(a, null, o));
```
meiliang86: Should supply the actual retry object here?

src/main/java/com/uber/cadence/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -735,8 +742,52 @@ private void fireTimer(String timerId) {
...
+                .setTaskStartToCloseTimeoutSeconds(startRequest.getTaskStartToCloseTimeoutSeconds())
+                .setTaskList(startRequest.getTaskList())
+                .setBackoffStartIntervalInSeconds(backoffIntervalSeconds)
+                .setRetryPolicy(startRequest.getRetryPolicy());
+        workflow.action(Action.CONTINUE_AS_NEW, ctx, continueAsNewAttr, decisionTaskCompletedId);
+        HistoryEvent event = ctx.getEvents().get(ctx.getEvents().size() - 1);
+        WorkflowExecutionContinuedAsNewEventAttributes continuedAsNewEventAttributes =
+            event.getWorkflowExecutionContinuedAsNewEventAttributes();
+        Optional<RetryState> continuedRetryState;
+        if (data.retryState.isPresent()) {
```
meiliang86: This is already checked in the outer if statement so it's guaranteed to exist.

src/main/java/com/uber/cadence/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -735,8 +742,52 @@ private void fireTimer(String timerId) {
...
+      String identity)
       throws InternalServiceError, BadRequestError {
+    WorkflowData data = workflow.getData();
+    if (data.retryState.isPresent()) {
+      int backoffIntervalSeconds =
+          data.retryState
+              .get()
+              .getBackoffIntervalInSeconds(d.getReason(), store.currentTimeMillis());
+      if (backoffIntervalSeconds > 0) {
+        ContinueAsNewWorkflowExecutionDecisionAttributes continueAsNewAttr =
```
meiliang86: Just want to confirm: Retry is implemented as ContinueAsNew only in the test framework, right? In real service it will be still the same run?

src/main/java/com/uber/cadence/internal/testservice/RetryState.java
```diff
@@ -0,0 +1,131 @@
...
+      return 0;
+    }
+
+    if (retryPolicy.getMaximumAttempts() > 0
+        && getAttempt() >= retryPolicy.getMaximumAttempts() - 1) {
+      // currAttempt starts from 0.
+      // MaximumAttempts is the total attempts, including initial (non-retry) attempt.
+      return 0;
+    }
+    long initInterval = retryPolicy.getInitialIntervalInSeconds() * MILLISECONDS_IN_SECOND;
```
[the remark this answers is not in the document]
mfateev: Completely forgot about it!

src/main/java/com/uber/cadence/internal/testservice/TestWorkflowMutableStateImpl.java
mfateev: No, the test framework matches the real service behavior.

PR #231, opened 2018-11-27
Child workflow server side retry

mfateev:
Any child workflow that used RetryOptions after this change is going to use the server side retry instead of the client side one. The change is backwards compatible.

Also added an ability to replay workflows from a JSON file or resource to test backwards compatibility.

### DIFF
--- a/src/main/java/com/uber/cadence/internal/testservice/TestWorkflowMutableStateImpl.java
+++ b/src/main/java/com/uber/cadence/internal/testservice/TestWorkflowMutableStateImpl.java
@@ -518,16 +518,49 @@ class TestWorkflowMutableStateImpl implements TestWorkflowMutableState {
       StartChildWorkflowExecutionDecisionAttributes a,
       long decisionTaskCompletedId)
       throws BadRequestError, InternalServiceError {
-    validateStartChildWorkflow(a);
+    validateStartChildExecutionAttributes(a);
     StateMachine<ChildWorkflowData> child = StateMachines.newChildWorkflowStateMachine(service);
     childWorkflows.put(ctx.getNextEventId(), child);
     child.action(StateMachines.Action.INITIATE, ctx, a, decisionTaskCompletedId);
     ctx.lockTimer();
   }
 
-  // TODO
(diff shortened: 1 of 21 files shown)


src/main/java/com/uber/cadence/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -525,7 +526,13 @@ private void processStartChildWorkflow(
   }
 
   // TODO
```
meiliang86: Remove this comment?

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -97,7 +98,7 @@
    * When set to true increases test, activity and workflow timeouts to large values to support
    * stepping through code in a debugger without timing out.
    */
-  private static final boolean DEBUGGER_TIMEOUTS = false;
+  private static final boolean DEBUGGER_TIMEOUTS = true;
```
meiliang86: Revert this change?

src/main/java/com/uber/cadence/internal/worker/WorkflowWorker.java
```diff
@@ -100,42 +101,58 @@ public void start() {
...
+    return queryWorkflowExecution(queryType, args, history, null);
+  }
+
+  private byte[] queryWorkflowExecution(
+      String queryType, byte[] args, WorkflowExecutionHistory history, byte[] nextPageToken)
+      throws Exception {
+    List<HistoryEvent> events = history.getEvents();
+    HistoryEvent startedEvent = events.get(0);
+    WorkflowExecutionStartedEventAttributes started =
+        startedEvent.getWorkflowExecutionStartedEventAttributes();
```
meiliang86: Move these three lines right before ```if (started == null) {``` (line 142)?

src/main/java/com/uber/cadence/internal/testservice/TestWorkflowMutableStateImpl.java
mfateev: Implemented validation.

PR #236, opened 2018-12-05
Added support for NonDeterministicWorkflowPolicy.FailWorkflow

mfateev:


src/main/java/com/uber/cadence/internal/replay/ReplayDecider.java
```diff
@@ -418,8 +420,17 @@ private void decideImpl(PollForDecisionTaskResponse decisionTask, Functions.Proc
         decisionsHelper.handleDecisionTaskStartedEvent(decision);
       }
     } catch (Error e) {
-      metricsScope.counter(MetricsType.DECISION_TASK_ERROR_COUNTER).inc(1);
-      throw e;
+      if (this.workflow.getWorkflowImplementationOptions().getNonDeterministicWorkflowPolicy()
```
meiliang86: Should we check if the error is NonDeterminisicWorkflowError?

mfateev: I think any Error should be treated this way. I kept the constant name similar to Go version for consistency.

PR #248, opened 2019-01-17
Add server side cron schedule support

meiliang86:


src/main/java/com/uber/cadence/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -829,6 +785,12 @@ private void processFailWorkflowExecution(
         return;
       }
     }
+
+    if (!Strings.isNullOrEmpty(data.cronSchedule)) {
+      startNewCronRun(ctx, decisionTaskCompletedId, identity, data, null);
```
mfateev: I believe service will use data from the last successful run

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -2887,6 +2889,65 @@ public void testWorkflowRetryWithMethodRetryDoNotRetryException() {
...
+
+      AtomicInteger count = retryCount.get(testName);
+      if (count == null) {
+        count = new AtomicInteger();
+        retryCount.put(testName, count);
+      }
+      int c = count.incrementAndGet();
+      SimpleDateFormat sdf = new SimpleDateFormat("MMM dd,yyyy HH:mm:ss.SSS");
+      Date now = new Date(Workflow.currentTimeMillis());
+      System.out.println("run at " + sdf.format(now));
```
mfateev: delete println or convert to log

PR #255, opened 2019-03-05
Expose methods to allow user to add headers in Thrift request

meiliang86:


build.gradle
```diff
@@ -37,7 +37,7 @@ googleJavaFormat {
 }
 
 group = 'com.uber.cadence'
-version = '2.3.0'
+version = '2.3.1'
```
mfateev: Would you update the release notes as well including my previous change as well?

meiliang86: Yes. Will update for both.

PR #257, opened 2019-03-21
Added unit testing against Cadence service running through docker

mfateev:


meiliang86:
This is done as part of #280

PR #264, opened 2019-04-08
Cron child workflow

meiliang86:


src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -3031,6 +3031,44 @@ public void testTestWorkflowWithCronSchedule() {
...
+
+    startWorkerFor(TestCronParentWorkflow.class, TestWorkflowWithCronScheduleImpl.class);
+
+    WorkflowStub client =
+        workflowClient.newUntypedWorkflowStub(
+            "TestWorkflow1::execute",
+            newWorkflowOptionsBuilder(taskList)
+                .setExecutionStartToCloseTimeout(Duration.ofHours(10))
+                .build());
+    registerDelayedCallback(Duration.ofHours(3), client::cancel);
```
mfateev: I prefer sleep after the start instead of the callback.

meiliang86: Done.

PR #272, opened 2019-04-24
Fix deadlock in sticky decider cache eviction

meiliang86:


src/main/java/com/uber/cadence/internal/sync/WorkflowThreadImpl.java
```diff
@@ -236,31 +236,25 @@ public void start() {
...
 
-    try {
-      taskFuture = threadPool.submit(task);
-      return;
-    } catch (RejectedExecutionException e) {
-      getDecisionContext()
-          .getMetricsScope()
-          .counter(MetricsType.STICKY_CACHE_THREAD_FORCED_EVICTION)
-          .inc(1);
+    while (true) {
```
mfateev: Should we have some time limit, just to be on the safe side?

meiliang86: We break if there's no item in the cache that can be evicted. Seems to me it's safe to always retry if we have a successful eviction.

PR #278, opened 2019-04-29
Allow data converter to handle non-serializable throwables in the cause chain

meiliang86:


src/main/java/com/uber/cadence/converter/CustomThrowableTypeAdapter.java
```diff
@@ -74,6 +94,11 @@ public void write(JsonWriter jsonWriter, T throwable) throws IOException {
       object.add("class", new JsonPrimitive(throwable.getClass().getName()));
       object.add("stackTrace", new JsonPrimitive(sw.toString()));
     } catch (Throwable e) {
+      // In case a throwable is not serializable, we will convert it to a data converter exception.
+      // The cause of the data converter exception will indicate why the serialization failed. On
+      // the
```
mfateev: weird comment formatting

meiliang86: Yeah the formatter changed the original text. I'll fix this.

PR #290, opened 2019-05-01
Remove migrated docs

dmetzgar:
The documentation that's removed from this readme is being moved to the central documentation site as part of https://github.com/uber/cadence/pull/1762.


dmetzgar:
Fixes #290

PR #292, opened 2019-05-13
Add missing metrics tags for activity and decision task processing

meiliang86:


src/main/java/com/uber/cadence/internal/worker/ActivityWorker.java
```diff
@@ -157,26 +160,31 @@ private TaskHandlerImpl(ActivityTaskHandler handler) {
 
     @Override
     public void handle(MeasurableActivityTask task) throws Exception {
-      options
-          .getMetricsScope()
+      Scope metricsScope =
+          options
+              .getMetricsScope()
+              .tagged(
+                  ImmutableMap.of(MetricsTag.ACTIVITY_TYPE, task.task.getActivityType().getName()));
```
mfateev: Why is it not using the activity type LoggerTag on the line 176?

meiliang86: Metrics and logging tag names are different.

shail: Can we also add the WORKFLOW_TYPE metrics tag?

meiliang86: Activity worker does not have that info. Workflow worker has that in decision task completion.

PR #298, opened 2019-05-17
Support Memo in visibility

vancexu:


src/main/java/com/uber/cadence/client/WorkflowOptions.java
```diff
@@ -181,6 +186,12 @@ public Builder setCronSchedule(String cronSchedule) {
       return this;
     }
 
+    /** Specifies additional information in result of list workflow. */
```
mfateev: additional non indexable information

src/main/java/com/uber/cadence/client/WorkflowOptions.java
```diff
@@ -356,6 +379,9 @@ public String toString() {
         + ", cronSchedule='"
         + cronSchedule
         + '\''
+        + ", memo='"
+        + memo.toString()
```
mfateev: What if memo is null?
In Java no need to call .toString when summing up a string. Just change to "memo="  + memo +

src/main/java/com/uber/cadence/internal/common/StartWorkflowExecutionParameters.java
```diff
@@ -363,6 +374,9 @@ public String toString() {
         + ", cronSchedule='"
         + cronSchedule
         + '\''
+        + ", memo='"
+        + memo.toString()
```
mfateev: ditto

src/main/java/com/uber/cadence/internal/sync/WorkflowStubImpl.java
```diff
@@ -136,9 +138,21 @@ private StartWorkflowExecutionParameters getStartWorkflowExecutionParameters(
...
     return p;
   }
 
+  private Map<String, byte[]> convertMemoFromObjectToBytes(Map<String, Object> memoFromOption) {
+    if (memoFromOption == null) {
+      return null;
+    }
+    Map<String, byte[]> memo = new HashMap<>();
+    for (Map.Entry<String, Object> item : memoFromOption.entrySet()) {
+      memo.put(item.getKey(), dataConverter.toData(item.getValue()));
```
mfateev: I would catch the exception and add information about which field of the memo failed serialization.

src/main/java/com/uber/cadence/workflow/ChildWorkflowOptions.java
```diff
@@ -346,6 +369,9 @@ public String toString() {
         + childPolicy
         + ", cronSchedule="
         + cronSchedule
+        + ", memo='"
+        + memo.toString()
```
mfateev: same

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -1147,10 +1147,17 @@ private void waitForProc(WorkflowExecution execution) {
...
+    memo.put("testKey", "testObject");
+    return memo;
+  }
+
   @Test
   public void testStart() {
     startWorkerFor(TestMultiargsWorkflowsImpl.class);
-    WorkflowOptions workflowOptions = newWorkflowOptionsBuilder(taskList).build();
+    WorkflowOptions workflowOptions =
+        newWorkflowOptionsBuilder(taskList).setMemo(getTestMemo()).build();
```
mfateev: Add check (may be using the WorkflowExecutionStarted event) that the memo was actually passed to the service.

src/main/java/com/uber/cadence/client/WorkflowOptions.java
vancexu: Updated, thanks

src/main/java/com/uber/cadence/internal/sync/WorkflowStubImpl.java
vancexu: updated

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
vancexu: Done

PR #306, opened 2019-05-24
Add some more fields in activity info

meiliang86:


src/main/java/com/uber/cadence/internal/sync/TestActivityEnvironmentInternal.java
```diff
@@ -208,7 +208,11 @@ public void setWorkflowService(IWorkflowService workflowService) {
       IWorkflowService service = new WorkflowServiceWrapper(workflowService);
       Result taskResult =
           activityTaskHandler.handle(
-              service, testEnvironmentOptions.getDomain(), task, NoopScope.getInstance());
+              service,
+              testEnvironmentOptions.getDomain(),
+              task,
+              NoopScope.getInstance(),
+              "unit-test");
```
mfateev: Use ActivityOptions.getTaskList()

src/main/java/com/uber/cadence/internal/worker/ActivityTaskHandler.java
```diff
@@ -84,7 +84,8 @@ Result handle(
       IWorkflowService service,
       String domain,
       PollForActivityTaskResponse activityTask,
-      Scope metricsScope);
+      Scope metricsScope,
+      String taskList);
```
mfateev: I would put taskList argument it after domain argument.

PR #312, opened 2019-05-30
Make NoopScope metric scope really a no-op operation

pgoron:
When workflow client code don't provide a metric scope, a default "NoopScope"
is instantiated by cadence client framework. With the current implementation of
NoopScope, a single threaded executor is indirectly created by RootScopeBuilder
for reporting empty metrics at regular interval. This executor is never shutdown.

The side effect is that a client process that just launches a workflow and returns
from main method will stay alive due to the non daemon thread of this executor.

Instead of tracking NoopScope instance created by cadence framework and to
call close() on it at the right time, it seemed easier to just rewrite the class to be
really no-op. In addition, it will save one thread when application don't provide
a metric scope.


pgoron:
closing/reopening PR to retrigger travis

Issue #267, opened 2019-04-16
Retry Options requires Expiration In Unit tests


pavankris:
Cadence Version : 2.3.2

If retry options does not have expiration , even though it has maximum attempts set. It fails 

`java.lang.RuntimeException: Failure processing decision task. WorkflowID=3d8091a8-8ac4-49b6-8d62-2ea4f4fb91e7, RunID=b47c211c-26e1-44be-a740-7175dc6bd542
    at com.uber.cadence.internal.worker.WorkflowWorker$TaskHandlerImpl.wrapFailure(WorkflowWorker.java:259) ~[cadence-client-2.3.1.jar:na]
    at com.uber.cadence.internal.worker.WorkflowWorker$TaskHandlerImpl.wrapFailure(WorkflowWorker.java:220) ~[cadence-client-2.3.1.jar:na]
    at com.uber.cadence.internal.worker.PollTaskExecutor.lambda$process$0(PollTaskExecutor.java:76) ~[cadence-client-2.3.1.jar:na]
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) ~[na:1.8.0_191]
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) ~[na:1.8.0_191]
    at java.lang.Thread.run(Thread.java:748) ~[na:1.8.0_191]
Caused by: com.uber.cadence.BadRequestError: ExpirationIntervalInSeconds must be greater than 0 on retry policy.`

`ActivityOptions ao =
          new ActivityOptions.Builder()
              .setScheduleToCloseTimeout(Duration.ofSeconds(90))
              .setTaskList(hostWorker)
              .setRetryOptions(
                  new RetryOptions.Builder()
                      .setInitialInterval(Duration.ofSeconds(1))
                      .setMaximumAttempts(2)
                      .build())
              .build();`
This was the activity options which was used

PR #315, opened 2019-05-31
Fix retry option without expiration in test env

meiliang86:
We allow user to specify either expiration or maxAttempt in retry options.
Currently if only maxAttempt is set, test Env state machine will throw exception.
Moreover, if only maxAttempt is set, expiration should be treated as infinite when deciding whether to retry or not.


meiliang86:
resolves #267

PR #318, opened 2019-06-07
Make sure signals are applied in the same order as they appear in history

meiliang86:


src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -4562,6 +4562,58 @@ public void testParallelLocalActivityExecutionWorkflow() {
         result);
   }
 
+  private static CompletableFuture<Boolean> signalSentForSignalOrderingWorkflowTest;
```
mfateev: Could you refactor test to not rely on global variable. It might break in the future when running multiple of them in parallel. One way to refactor is to complete the workflow on a specific signal value.

PR #323, opened 2019-06-14
Default retry policy for Cadence service API calls

### LINKED ISSUE #258: Dynamic retry policy for Cadence service API calls
Without such policy errors can be received during service deployments. Correspondent Go Client PR:

https://github.com/uber-go/cadence-client/pull/668#pullrequestreview-219743378

meiliang86:
resolves #258
In Java we don't have deadline in context, so we only have the default retry policy, and do not adjust based on context.deadline as Go does.

### DIFF
--- a/src/main/java/com/uber/cadence/common/RetryOptions.java
+++ b/src/main/java/com/uber/cadence/common/RetryOptions.java
@@ -18,13 +18,6 @@
 package com.uber.cadence.common;
 
 import com.google.common.base.Defaults;
-import com.uber.cadence.BadRequestError;
-import com.uber.cadence.CancellationAlreadyRequestedError;
-import com.uber.cadence.DomainAlreadyExistsError;
-import com.uber.cadence.DomainNotActiveError;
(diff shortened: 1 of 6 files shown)


src/main/java/com/uber/cadence/common/RetryOptions.java
```diff
@@ -27,10 +34,37 @@
 import java.util.Objects;
 
 public final class RetryOptions {
+  public static final RetryOptions DEFAULT_SERVICE_OPERATION_RETRY_OPTIONS;
 
+  private static final Duration RETRY_SERVICE_OPERATION_INITIAL_INTERVAL = Duration.ofMillis(20);
+  private static final Duration RETRY_SERVICE_OPERATION_EXPIRATION_INTERVAL = Duration.ofMinutes(1);
+  private static final double RETRY_SERVICE_OPERATION_BACKOFF = 1.2;
```
mfateev: I don't think these constants belong here. The retry options are used by workflow code that shouldn't really be aware about underlying RPC layer.

PR #324, opened 2019-06-15
Change markers to use headers to serialize internal fields

### LINKED ISSUE #305: Update SideEffect and MutableSideEffect to use headers
Currently side effect markers serialize a structure that contains a serialized payload into data field. This leads to the following JSON as Marker details:

```
 5: MarkerRecorded [191 ms] {
    Details = {
      "id": "c661391c-feaa-3148-bcb8-a2c401953b50",
      "eventId": 5,
      "data": [
        123,
        34,
   ...
        34,
        58,
        110,
        117,
        108,
        108,
        125
      ],
      "accessCount": 0
    };
    DecisionTaskCompletedEventId = 3;
    MarkerName = MutableSideEffect
  },
```
The proposal is to use header field of the Marker to store id, eventId and accessCount and only put data into the Details.

meiliang86:
resolves #305


src/main/java/com/uber/cadence/internal/common/LocalActivityMarkerData.java
```diff
@@ -85,77 +92,115 @@ public LocalActivityMarkerData build() {
...
+    return header;
+  }
+
+  public static LocalActivityMarkerData fromEventAttributes(
+      MarkerRecordedEventAttributes attributes) {
+    ByteBuffer byteBuffer = attributes.getHeader().getFields().get(LOCAL_ACTIVITY_HEADER_KEY);
+    byte[] bytes = org.apache.thrift.TBaseHelper.byteBufferToByteArray(byteBuffer);
+    LocalActivityMarkerHeader header =
+        JsonDataConverter.getInstance()
+            .fromData(bytes, LocalActivityMarkerHeader.class, LocalActivityMarkerHeader.class);
```
mfateev: We shouldn't hardcode the converter used to serialize/deserialize the header. The converter should be passed from the higher level code, probably as a parameter to fromEventAttributes and getHeader.

src/main/java/com/uber/cadence/internal/replay/DecisionsHelper.java
```diff
@@ -17,41 +17,7 @@
...
-import com.uber.cadence.StartChildWorkflowExecutionDecisionAttributes;
-import com.uber.cadence.StartChildWorkflowExecutionFailedEventAttributes;
-import com.uber.cadence.StartChildWorkflowExecutionInitiatedEventAttributes;
-import com.uber.cadence.StartTimerDecisionAttributes;
-import com.uber.cadence.TaskList;
-import com.uber.cadence.TimerCanceledEventAttributes;
-import com.uber.cadence.TimerFiredEventAttributes;
-import com.uber.cadence.WorkflowExecutionStartedEventAttributes;
-import com.uber.cadence.WorkflowType;
+import com.uber.cadence.*;
```
mfateev: revert. There is IntelliJ setting to avoid this.

meiliang86: Done

src/main/java/com/uber/cadence/internal/common/LocalActivityMarkerData.java
meiliang86: Done

src/main/java/com/uber/cadence/internal/common/LocalActivityMarkerData.java
```diff
@@ -85,77 +93,114 @@ public LocalActivityMarkerData build() {
...
+    return headers.backoff;
   }
 
   public boolean getIsCancelled() {
-    return isCancelled;
+    return headers.isCancelled;
+  }
+
+  public Header getHeader() {
+    byte[] headerData = JsonDataConverter.getInstance().toData(headers);
```
mfateev: What about here?

src/main/java/com/uber/cadence/internal/replay/ClockDecisionContext.java
```diff
@@ -88,20 +86,23 @@ public void accept(Exception reason) {
...
   private final Lock laTaskLock = new ReentrantLock();
   private final Condition taskCondition = laTaskLock.newCondition();
   private boolean taskCompleted = false;
 
   ClockDecisionContext(
       DecisionsHelper decisions,
       BiFunction<LocalActivityWorker.Task, Duration, Boolean> laTaskPoller,
-      ReplayDecider replayDecider) {
+      ReplayDecider replayDecider,
+      DataConverter dataConverter) {
```
mfateev: I believe DataConverter should live in higher level. At this level everything is a byte. Look at getVersion call in this file.

meiliang86: The difference is getVersion is from workflow code, so we can access data converter from SyncDecisionContext and pass it down.
Here deserialization needs to happen while we replay event. So we will need to wire data converter into replay decider.

src/main/java/com/uber/cadence/internal/common/LocalActivityMarkerData.java
meiliang86: Fixed.

PR #325, opened 2019-06-19
Lock decider while processing

meiliang86:
This prevents query from messing up decision tasks in progress, especially when there's local activity.
Query should only do replay of local activities, not executing additional ones.
Also Start decision task should not look at sticky schedule to start timeout.

### DIFF
--- a/src/main/java/com/uber/cadence/internal/replay/ReplayDecider.java
+++ b/src/main/java/com/uber/cadence/internal/replay/ReplayDecider.java
@@ -51,6 +51,8 @@ import java.util.Objects;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import org.apache.thrift.TException;
@@ -75,6 +77,7 @@ class ReplayDecider implements Decider, Consumer<HistoryEvent> {
   private final Scope metricsScope;
   private final long wfStartTimeNanos;
   private final WorkflowExecutionStartedEventAttributes startedEvent;
+  private Lock inProcessing = new ReentrantLock();
 
   ReplayDecider(
       IWorkflowService service,
@@ -353,8 +356,14 @@ class ReplayDecider implements Decider, Consumer<HistoryEvent> {
 
   @Override
   public DecisionResult decide(PollForDecisionTaskResponse decisionTask) throws Throwable {
-    boolean forceCreateNewDecisionTask = decideImpl(decisionTask, null);
-    return new DecisionResult(decisionsHelper.getDecisions(), forceCreateNewDecisionTask);
+    inProcessing.lock();
+
+    try {
+      boolean forceCreateNewDecisionTask = decideImpl(decisionTask, null);
+      return new DecisionResult(decisionsHelper.getDecisions(), forceCreateNewDecisionTask);
+    } finally {
+      inProcessing.unlock();
+    }
   }
 
   // Returns boolean to indicate whether we need to force create new decision task for local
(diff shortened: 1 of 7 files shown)


src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -4583,6 +4585,74 @@ public void testParallelLocalActivityExecutionWorkflow() {
...
+    }
+  }
+
+  @Test
+  public void testLocalActivityAndQuery() throws InterruptedException {
+    startWorkerFor(TestLocalActivityAndQueryWorkflow.class);
+    WorkflowOptions options =
+        new WorkflowOptions.Builder()
+            .setExecutionStartToCloseTimeout(Duration.ofMinutes(5))
+            .setTaskStartToCloseTimeout(Duration.ofSeconds(10))
```
mfateev: set start to close timeout to 1 second to force immediate decision

src/main/java/com/uber/cadence/internal/replay/ReplayDecider.java
```diff
@@ -353,8 +356,14 @@ private void handleWorkflowExecutionSignaled(HistoryEvent event) {
 
   @Override
   public DecisionResult decide(PollForDecisionTaskResponse decisionTask) throws Throwable {
-    boolean forceCreateNewDecisionTask = decideImpl(decisionTask, null);
-    return new DecisionResult(decisionsHelper.getDecisions(), forceCreateNewDecisionTask);
+    inProcessing.lock();
```
mfateev: I don't think it is going to work with immediate decision.

meiliang86: We have discussed and decided that it's fine to process query during local activity batches, as processed local activities are processed and it will not cause non-determinism for queries.

src/main/java/com/uber/cadence/internal/worker/WorkflowRunLockManager.java
```diff
@@ -0,0 +1,102 @@
...
+
+  private final ReentrantLock mapLock = new ReentrantLock();
+  private final HashMap<String, CountableLock> perRunLock = new HashMap<>();
+
+  ReentrantLock getLockForLocking(String runId) {
+    mapLock.lock();
+
+    try {
+      CountableLock cl;
+      if (perRunLock.containsKey(runId)) {
```
mfateev: In java call get and then check if it is null. contains then get makes double lookup.

src/test/java/com/uber/cadence/internal/worker/WorkflowRunLockManagerTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class WorkflowRunLockManagerTest {
+  private static final Logger log = LoggerFactory.getLogger(WorkflowRunLockManagerTest.class);
+  private WorkflowRunLockManager runLockManager = new WorkflowRunLockManager();
+
+  @Test
+  public void lockAndUnlockTest() throws ExecutionException, InterruptedException {
```
mfateev: to be complete add one more key

src/main/java/com/uber/cadence/internal/worker/WorkflowRunLockManager.java
```diff
@@ -43,25 +39,28 @@ int getCount() {
...
     }
   }
 
-  private final ReentrantLock mapLock = new ReentrantLock();
+  private final Lock mapLock = new ReentrantLock();
   private final HashMap<String, CountableLock> perRunLock = new HashMap<>();
 
-  ReentrantLock getLockForLocking(String runId) {
+  // This method returns a lock that can be used to serialize decision task processing for a
+  // particular workflow
```
mfateev: Make it Javadoc comment

Issue #316, opened 2019-06-05
Expose cron scheduled time though workflow.GetInfo


mfateev:
https://github.com/uber-go/cadence-client/issues/738

mfateev:
Also add workflow start time and other fields recently added to DecisionTaskStarted event:

https://github.com/uber-go/cadence-client/issues/750

PR #333, opened 2019-07-02
Expose wf schedule/execution time 

meiliang86:
Also fix cron first execution time in test framework


meiliang86:
#316

src/main/java/com/uber/cadence/internal/replay/WorkflowContext.java
```diff
@@ -131,4 +136,14 @@ void setCurrentRunId(String currentRunId) {
...
   }
+
+  long getScheduleTimeMillis() {
+    return TimeUnit.NANOSECONDS.toMillis(
+        decisionTask.getHistory().getEvents().get(0).getTimestamp());
+  }
+
+  long getExecutionTimeMillis() {
+    return TimeUnit.NANOSECONDS.toMillis(
+        decisionTask.getHistory().getEvents().get(1).getTimestamp());
```
mfateev: I don't think it is always true. For example signalWithStart will have signal as a second event. 
Also when cron workflow is waiting for exectution all received signals will be put before DecisionTaskScheduled.

mfateev (changes requested):

shijiesheng:

PR #334, opened 2019-07-08
Wrap data converter exception in activity failure exception

meiliang86:
Currently only adding activity type. ActivityId and EventId is not available right now in SyncDecisionContext and it's not trivial to add. My understanding is that activity type should be good enough for trouble shooting purpose. Open to discussions though.


mfateev (commented):
Add unit test

shijiesheng:

Issue #294, opened 2019-05-15
Add protobuf support to the default DataConverter


mfateev:

xsilmarx:
Hi, I see the PR is still open for almost one year, but it seems is only dependent on a small change. Is there a ETA for this PR to be merged? Thanks!

PR #335, opened 2019-07-08
Add support for protobuf in default data converter

meiliang86:


meiliang86:
#294

src/main/java/com/uber/cadence/converter/JsonDataConverter.java
```diff
@@ -88,6 +90,11 @@ public JsonDataConverter(Function<GsonBuilder, GsonBuilder> builderInterceptor)
     try {
       if (values.length == 1) {
         Object value = values[0];
+        if (value instanceof GeneratedMessageV3) {
+          GeneratedMessageV3 protoObj = (GeneratedMessageV3) value;
+          return protoObj.toByteArray();
```
mfateev: I think by default we want to serialize it to JSON to make it UI friendly: https://github.com/protocolbuffers/protobuf/blob/master/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java

mfateev (changes requested):

shijiesheng:

Issue #336, opened 2019-07-11
RPC Timeout warning on long-running workflows


khaale:
Java Client v 2.5.0 shows a warning when running a longer than 1sec workflow. 

Steps to reproduce:
* Use latest compose file from cadence/docker
* Download java-client-samples
* Update java-client version to **2.5.0**
* Put `Wokflow.sleep(1000)` somewhere in a sample workflow body.
* Run sample

Expected results: the same as without sleep.
Actual results: warning message
```
21:38:21.691 [main] WARN c.u.cadence.internal.common.Retryer - Retrying after failure
org.apache.thrift.TException: Rpc error:<ErrorResponse id=2 errorType=Timeout message=timeout>
at com.uber.cadence.serviceclient.WorkflowServiceTChannel.throwOnRpcError(WorkflowServiceTChannel.java:505)
at com.uber.cadence.serviceclient.WorkflowServiceTChannel.doRemoteCall(WorkflowServiceTChannel.java:480)
at com.uber.cadence.serviceclient.WorkflowServiceTChannel.getWorkflowExecutionHistory(WorkflowServiceTChannel.java:815)
at com.uber.cadence.serviceclient.WorkflowServiceTChannel.lambda$GetWorkflowExecutionHistory$8(WorkflowServiceTChannel.java:804)
at com.uber.cadence.serviceclient.WorkflowServiceTChannel.measureRemoteCall(WorkflowServiceTChannel.java:525)
at com.uber.cadence.serviceclient.WorkflowServiceTChannel.GetWorkflowExecutionHistory(WorkflowServiceTChannel.java:803)
at com.uber.cadence.internal.common.WorkflowExecutionUtils.lambda$getInstanceCloseEvent$1(WorkflowExecutionUtils.java:201)
at com.uber.cadence.internal.common.Retryer.retryWithResult(Retryer.java:122)
at com.uber.cadence.internal.common.WorkflowExecutionUtils.getInstanceCloseEvent(WorkflowExecutionUtils.java:201)
at com.uber.cadence.internal.common.WorkflowExecutionUtils.getWorkflowExecutionResult(WorkflowExecutionUtils.java:127)
at com.uber.cadence.internal.sync.WorkflowStubImpl.getResult(WorkflowStubImpl.java:231)
at com.uber.cadence.internal.sync.WorkflowStubImpl.getResult(WorkflowStubImpl.java:213)
[...]

PR #337, opened 2019-07-11
Fix busy loop in local activity poller if there is no task

meiliang86:
Also fix an issue in get history timeout.


meiliang86:
#336

meiliang86:
#338

src/main/java/com/uber/cadence/internal/worker/LocalActivityPollTask.java
```diff
@@ -33,7 +33,11 @@
 
   @Override
   public LocalActivityWorker.Task poll() throws TException {
-    return pendingTasks.poll();
+    try {
+      return pendingTasks.take();
+    } catch (InterruptedException e) {
+      throw new RuntimeException("local activity poll task interrupted");
```
mfateev: Chain e just for completeness

PR #340, opened 2019-07-15
Add saga class that helps user to implement saga pattern in Cadence

meiliang86:


src/main/java/com/uber/cadence/workflow/Saga.java
```diff
@@ -0,0 +1,37 @@
...
+  public Saga(boolean parallelCompensation) {
+    this.parallelCompensation = parallelCompensation;
+  }
+
+  public void addCompensation(Functions.Proc func) {
+    compensations.addFirst(func);
+  }
+
+
+  public void compensate() {
```
meiliang86: What should we do if compensation fails?

mfateev: Have an option to either ignore errors or fail.

src/main/java/com/uber/cadence/workflow/Saga.java
```diff
@@ -0,0 +1,186 @@
...
+ *
+ *  or in the "license" file accompanying this file. This file is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package com.uber.cadence.workflow;
+
+import com.uber.cadence.client.BatchRequest;
```
mfateev: Let's rename it to OperationsBatch (or you have better idea?).
And move to common package as client package doesn't make sense here.

src/main/java/com/uber/cadence/workflow/Saga.java
```diff
@@ -0,0 +1,186 @@
...
+        results.add(f.apply());
+      }
+
+      RuntimeException sagaException = null;
+      for (Promise p : results) {
+        try {
+          p.get();
+        } catch (Exception e) {
+          if (sagaException == null) {
+            sagaException = new RuntimeException("Exception from saga compensate", e);
```
mfateev: May be introduce SagaException?

src/main/java/com/uber/cadence/workflow/Saga.java
meiliang86: I decided to not implement the interface, just add the methods and call them addCompensate explicitly. Let me know what you think.

src/main/java/com/uber/cadence/workflow/Saga.java
khaale: With both `parallelCompensation` and `continueWithError` options, compensation behavior looks a bit implicit:

| parallelCompensation  | continueWithError | real behavior in case of failure |
| ------------- | ------------- |--------|
| false  | false  | as expected |
| false  | true | as expected |
| true  | false  | efficiently ignored? |
| true  | true | as expected |

src/main/java/com/uber/cadence/workflow/Saga.java
```diff
@@ -0,0 +1,313 @@
...
+          if (!options.continueWithError) {
+            throw e;
+          }
+        }
+      }
+    }
+  }
+
+  /**
+   * Add compensation operation for saga, which will be executed in the reverse order if {@link
```
khaale: `in the reverse order` - is it true in case of parallel compensation?

meiliang86: If it's parallel then there is no order...I'll call this out explicitly probably.

meiliang86: So you are right continueWithError is mostly useful when you don't use parallel compensation. If you choose to do parallel compensation, then all the compensation operations will be fired no matter what and then you will receive exceptions back if there's any. I'll update the java docs.

khaale: I'd just remove these words from addCompensation comments

src/main/java/com/uber/cadence/workflow/Saga.java
```diff
@@ -0,0 +1,330 @@
...
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package com.uber.cadence.workflow;
+
+import java.util.ArrayList;
+import java.util.List;
+
```
mfateev: Add comment with explanation and code sample.

src/main/java/com/uber/cadence/workflow/Saga.java
mfateev: I think it is more about reporting if error happened. When continueWithError is false all errors are ignored.

src/main/java/com/uber/cadence/workflow/Saga.java
```diff
@@ -0,0 +1,330 @@
...
+   * Saga#compensate()} is called.
+   *
+   * @param operation to be executed during compensation.
+   */
+  public void addCompensation(Functions.Proc operation) {
+    compensationOps.add(() -> Async.procedure(operation));
+  }
+
+  /**
+   * Add compensation operation for saga, which will be executed in the reverse order if {@link
```
mfateev: I think "in the reverse order" comments are incorrect when parallel is set.

src/main/java/com/uber/cadence/workflow/Saga.java
meiliang86: Yes will do.

PR #358, opened 2019-08-13
Sync IDLs and add new APIs

vancexu:
Sync IDLs for next feature work.
Add related implementation for new APIs.

Also update docker to reflect recent server side change.


src/main/java/com/uber/cadence/serviceclient/WorkflowServiceTChannel.java
```diff
@@ -1533,6 +1541,15 @@ public StartWorkflowExecutionResponse SignalWithStartWorkflowExecution(
         () -> signalWithStartWorkflowExecution(signalWithStartRequest));
   }
 
+  // TODO: https://github.com/uber/cadence-java-client/issues/359
```
meiliang86: Why is this needed in client?

PR #364, opened 2019-08-21
Support Search Attributes on Start workflow

vancexu:
https://github.com/uber/cadence-java-client/issues/347

### DIFF
--- a/src/main/java/com/uber/cadence/client/WorkflowOptions.java
+++ b/src/main/java/com/uber/cadence/client/WorkflowOptions.java
@@ -64,6 +64,7 @@ public final class WorkflowOptions {
         .setRetryOptions(RetryOptions.merge(methodRetry, o.getRetryOptions()))
         .setCronSchedule(OptionsUtils.merge(cronAnnotation, o.getCronSchedule(), String.class))
         .setMemo(o.getMemo())
+        .setSearchAttributes(o.getSearchAttributes())
         .validateBuildWithDefaults();
   }
 
@@ -87,6 +88,8 @@ public final class WorkflowOptions {
 
     private Map<String, Object> memo;
 
+    private Map<String, Object> searchAttributes;
+
     public Builder() {}
 
     public Builder(WorkflowOptions o) {
@@ -102,6 +105,7 @@ public final class WorkflowOptions {
       this.retryOptions = o.retryOptions;
       this.cronSchedule = o.cronSchedule;
       this.memo = o.memo;
+      this.searchAttributes = o.searchAttributes;
     }
 
     /**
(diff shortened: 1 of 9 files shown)


src/main/java/com/uber/cadence/client/WorkflowOptions.java
```diff
@@ -192,6 +196,12 @@ public Builder setMemo(Map<String, Object> memo) {
       return this;
     }
 
+    /** Specifies additional indexed information in result of list workflow. */
```
mfateev: Which types are allowed as values? Are they serialized in any way?

mfateev: The same question applies to memo.

src/main/java/com/uber/cadence/internal/sync/WorkflowStubImpl.java
```diff
@@ -139,24 +139,24 @@ private StartWorkflowExecutionParameters getStartWorkflowExecutionParameters(
     }
     p.setInput(dataConverter.toData(args));
     p.setWorkflowType(new WorkflowType().setName(workflowType.get()));
-    p.setMemo(convertMemoFromObjectToBytes(o.getMemo()));
+    p.setMemo(convertMapFromObjectToBytes(o.getMemo()));
+    p.setSearchAttributes(convertMapFromObjectToBytes(o.getSearchAttributes()));
```
mfateev: I don't think you want JSON as search attribute value.

src/test/java/com/uber/cadence/client/WorkflowOptionsTest.java
```diff
@@ -188,4 +196,10 @@ public void testInvalidCronScheduleAnnotation() throws NoSuchMethodException {
     memo.put("testKey", "testObject");
     return memo;
   }
+
+  private Map<String, Object> getTestSearchAttributes() {
+    Map<String, Object> searchAttr = new HashMap<>();
+    searchAttr.put("CustomKeywordField", "testKey");
```
mfateev: May be more interesting type than string?

src/main/java/com/uber/cadence/client/WorkflowOptions.java
vancexu: Comments updated

src/test/java/com/uber/cadence/client/WorkflowOptionsTest.java
vancexu: Added more types

src/main/java/com/uber/cadence/internal/sync/WorkflowStubImpl.java
vancexu: You are right, for now client-server requires such Json encoding, but we can remove it. https://github.com/uber/cadence/issues/2433

PR #377, opened 2019-09-04
CancellationScope improvements and documentation

mfateev:
Added `Workflow.newCancellationScope(Functions.Proc1<CancellationScope> proc)` to allow referencing the wrapping scope inside the wrapped code.

Added CancellationScope documentation.


src/main/java/com/uber/cadence/internal/sync/CancellationScopeImpl.java
```diff
@@ -73,6 +74,13 @@ private static void popCurrent(CancellationScopeImpl expected) {
     setParent(parent);
   }
 
+  public CancellationScopeImpl(
+      boolean ignoreParentCancellation, Functions.Proc1<CancellationScope> proc) {
+    this.detached = ignoreParentCancellation;
+    this.runnable = () -> proc.apply(this);
+    setParent(current());
```
samarabbas: Do need another constructor to pass parent explicitly like other constructor?

src/main/java/com/uber/cadence/workflow/Workflow.java
```diff
@@ -589,7 +605,52 @@ public static CancellationScope newCancellationScope(Runnable runnable) {
...
+   *                        });
+   *                     Promise.allOf(p1, p2).get();
+   *                   })
+   *               .run();
+   * </code></pre>
+   *
+   * @param proc code to wrap in the cancellation scope
+   * @return wrapped proc
+   */
+  public static CancellationScope newCancellationScope(Functions.Proc1<CancellationScope> proc) {
```
samarabbas: Do you think another API to create newDetachedCancellationScope makes sense which takes parent?

src/main/java/com/uber/cadence/internal/sync/CancellationScopeImpl.java
mfateev: The only place where the parent is used is when a new child thread is created and the thread cancellation scope should be connected to the parents thread cancellation scope. So no need for the parent parameter when Proc1 one is passed.

src/main/java/com/uber/cadence/workflow/Workflow.java
mfateev: It might theoretically. I don't envision any specific use case when it might be useful. Let's postpone its creation until a real use case comes up.

Issue #379, opened 2019-09-05
Add missing metrics from go client


vancexu:
Need to find all missing metrics from go client and add them, such as `activity-scheduled-to-start-latency`

aai95:
Hello @vancexu
I have signed Uber CLA.
Could you please assign this issue to me?
Also, the target branch for my PR is master, right?

aai95:
@vancexu
As I understand, the following Go metrics have their semantic copy in Java:
Go `decision-task-panic` has name in Java `decision-task-error`
Go `decision-task-force-completed` **should be added**
Go `activity-task-panic` has name in Java `activity-task-error`
Go `sticky-cache-evict` has names in Java `sticky-cache-total-forced-eviction` and `sticky-cache-thread-forced-eviction`
Go `non-deterministic-error` **should be added**

I can't find proper place in Java client for adding of this two metrics.
Do we already have required functionality implementation for injection of this two metrics?

Thank you in advance for your assistance!

vancexu:
> Hello @vancexu
> I have signed Uber CLA.
> Could you please assign this issue to me?
> Also, the target branch for my PR is master, right?

Thank you for the contribution! 
Yes, target branch is master

PR #380, opened 2019-09-06
Add metrics for SCHEDULED_TO_STAR latency

vancexu:
https://github.com/uber/cadence-java-client/issues/379

### DIFF
--- a/src/main/java/com/uber/cadence/internal/worker/ActivityPollTask.java
+++ b/src/main/java/com/uber/cadence/internal/worker/ActivityPollTask.java
@@ -26,6 +26,7 @@ import com.uber.cadence.TaskListMetadata;
 import com.uber.cadence.internal.metrics.MetricsType;
 import com.uber.cadence.serviceclient.IWorkflowService;
 import com.uber.m3.tally.Stopwatch;
+import com.uber.m3.util.Duration;
 import org.apache.thrift.TException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


src/main/java/com/uber/cadence/internal/worker/ActivityPollTask.java
```diff
@@ -91,6 +92,10 @@ public ActivityPollTask(
     }
 
     options.getMetricsScope().counter(MetricsType.ACTIVITY_POLL_SUCCEED_COUNTER).inc(1);
+    options
+        .getMetricsScope()
+        .timer(MetricsType.ACTIVITY_SCHEDULED_TO_START_LATENCY)
+        .record(Duration.ofNanos(System.nanoTime() - result.scheduledTimestamp));
```
mfateev: result.getStartedTimestamp - result.getScheduledTimestampOfThisAttempt

src/main/java/com/uber/cadence/internal/worker/WorkflowPollTask.java
```diff
@@ -101,6 +102,9 @@ public PollForDecisionTaskResponse poll() throws TException {
     }
 
     metricScope.counter(MetricsType.DECISION_POLL_SUCCEED_COUNTER).inc(1);
+    metricScope
+        .timer(MetricsType.DECISION_SCHEDULED_TO_START_LATENCY)
+        .record(Duration.ofNanos(System.nanoTime() - result.scheduledTimestamp));
```
mfateev: ditto

src/main/java/com/uber/cadence/internal/worker/ActivityPollTask.java
vancexu: addressed. also fixed go client in https://github.com/uber-go/cadence-client/pull/836

src/main/java/com/uber/cadence/internal/worker/WorkflowPollTask.java
vancexu: updated (but use getScheduledTimestamp because decision doesn't have getScheduledTimestampOfThisAttempt)

PR #405, opened 2019-10-10
Cleanup threads created in signal method on decider close

meiliang86:


src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -5194,6 +5194,59 @@ public void testSagaParallelCompensation() {
...
+  @Test
+  public void testExceptionInSignal() throws InterruptedException {
+    startWorkerFor(TestSignalExceptionWorkflowImpl.class);
+    TestWorkflowSignaled signalWorkflow = workflowClient.newWorkflowStub(
+        TestWorkflowSignaled.class, newWorkflowOptionsBuilder(taskList).build());
+    CompletableFuture<String> result = WorkflowClient.execute(signalWorkflow::execute);
+    signalWorkflow.signal1("test");
+    try {
+      result.get(1, TimeUnit.SECONDS);
+    } catch (Exception e) {
```
mfateev: add fail("not reachable") before catch

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -5194,6 +5194,59 @@ public void testSagaParallelCompensation() {
...
+
+    // Suspend polling so that decision tasks are not retried. Otherwise it will affect our thread count.
+    if (useExternalService) {
+      workerFactory.suspendPolling();
+    } else {
+      testEnvironment.getWorkerFactory().suspendPolling();
+    }
+
+    // Wait for decision task retry to finish.
+    Thread.sleep(10000);
```
mfateev: Why do we retry decision task on exception? We should retry it only on Error.

meiliang86: Because we want user to fix the code and then the execution can continue, right?

meiliang86: Took a closer look, it seems that signal thread/handling is quite different than workflow thread... So here we are not catching the Exception and convert it to WorkflowExecutionException.

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
meiliang86: done

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
meiliang86: So the behavior is different between signal method and workflow method, but this can be a separate issue though...

mfateev: OK. Would you file an issue to review the signal method error handling?

meiliang86:
https://github.com/uber/cadence-java-client/issues/407

PR #406, opened 2019-10-10
Add Upsert Search Attributes

vancexu:
This PR add new feature UpsertSearchAttributes to allow use update search attributes inside workflow implementation. 

TODO in separate PR is add search attributes to workflow info, so that user can easily get such info.

### DIFF
--- a/src/main/java/com/uber/cadence/internal/replay/ClockDecisionContext.java
+++ b/src/main/java/com/uber/cadence/internal/replay/ClockDecisionContext.java
@@ -222,7 +222,7 @@ public final class ClockDecisionContext {
     decisions.upsertSearchAttributes(searchAttributes);
   }
 
-  void handleUpsertSearchAttributes(HistoryEvent event) {
+  void handleUpsertSearchAttributesEvent(HistoryEvent event) {
     // todo: update workflow info
   }
 
--- a/src/main/java/com/uber/cadence/internal/replay/ClockDecisionContext.java
+++ b/src/main/java/com/uber/cadence/internal/replay/ClockDecisionContext.java
@@ -21,6 +21,7 @@ import com.google.common.base.Strings;
 import com.uber.cadence.ActivityType;
 import com.uber.cadence.HistoryEvent;
 import com.uber.cadence.MarkerRecordedEventAttributes;
+import com.uber.cadence.SearchAttributes;
 import com.uber.cadence.StartTimerDecisionAttributes;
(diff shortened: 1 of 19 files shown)


src/main/java/com/uber/cadence/workflow/Workflow.java
```diff
@@ -1182,6 +1183,39 @@ public static Logger getLogger(String name) {
...
+   *
+   * <pre><code>
+   *     Map<String, Object> attr1 = new HashMap<>();
+   *     attr1.put("CustomIntField", 1);
+   *     attr1.put("CustomBoolField", true);
+   *     Workflow.upsertSearchAttributes(attr1);
+   *
+   *     Map<String, Object> attr2 = new HashMap<>();
+   *     attr2.put("CustomIntField", 2);
+   *     attr2.put("CustomKeywordField", "key");
```
meiliang86: value should be seattle?

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -183,7 +183,7 @@ protected void failed(Throwable e, Description description) {
         }
       };
 
-  @Parameter public boolean useExternalService;
+  @Parameter public boolean useExternalService = true;
```
meiliang86: revert this before checking in.

src/main/java/com/uber/cadence/internal/replay/ClockDecisionContext.java
```diff
@@ -217,6 +218,14 @@ private void timerCancelled(long startEventId, Exception reason) {
     return mutableSideEffectHandler.handle(id, converter, func);
   }
 
+  void upsertSearchAttributes(SearchAttributes searchAttributes) {
+    decisions.upsertSearchAttributes(searchAttributes);
+  }
+
+  void handleUpsertSearchAttributes(HistoryEvent event) {
```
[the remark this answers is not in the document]
vancexu: done

src/main/java/com/uber/cadence/workflow/Workflow.java
vancexu: yep, thanks

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
vancexu: thanks for the catch. done

vancexu:
> Have you updated the unit test framework?

Let me add test framework support in another PR.

PR #409, opened 2019-10-15
Support get search attributes inside workflow

vancexu:
This PR provides necessary APIs for users to get search attributes inside workflow. 

TODO: in next PR, I will add necessary test framework support.

### DIFF
--- a/src/main/java/com/uber/cadence/workflow/WorkflowInfo.java
+++ b/src/main/java/com/uber/cadence/workflow/WorkflowInfo.java
@@ -18,6 +18,7 @@
 package com.uber.cadence.workflow;
 
 import com.uber.cadence.ChildPolicy;
+import com.uber.cadence.SearchAttributes;
 import java.time.Duration;
 
 public interface WorkflowInfo {
@@ -35,4 +36,6 @@ public interface WorkflowInfo {
   Duration getExecutionStartToCloseTimeout();
 
   ChildPolicy getChildPolicy();
+
+  SearchAttributes getSearchAttributes();
 }
--- a/src/main/java/com/uber/cadence/workflow/WorkflowUtils.java
+++ b/src/main/java/com/uber/cadence/workflow/WorkflowUtils.java
@@ -1,0 +1,43 @@
+/*
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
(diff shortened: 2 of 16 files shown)


src/main/java/com/uber/cadence/workflow/WorkflowUtils.java
```diff
@@ -0,0 +1,43 @@
...
+import com.cronutils.utils.StringUtils;
+import com.uber.cadence.SearchAttributes;
+import com.uber.cadence.converter.DataConverter;
+import com.uber.cadence.converter.JsonDataConverter;
+import java.nio.ByteBuffer;
+
+public class WorkflowUtils {
+  private static final DataConverter jsonConverter = JsonDataConverter.getInstance();
+
+  public static <T> T getValueFromSearchAttributes(
```
meiliang86: IMO this should not be part of the library. Cadence library should only return byte array. User is responsible to translate that back to strongly typed object, with whatever data converter they prefer.

src/main/java/com/uber/cadence/workflow/WorkflowInfo.java
```diff
@@ -35,4 +36,6 @@
   Duration getExecutionStartToCloseTimeout();
 
   ChildPolicy getChildPolicy();
+
+  SearchAttributes getSearchAttributes();
```
meiliang86: This should not be the thrift struct, but the actual object?

meiliang86: Cadence knows the DataConverter that we use to serialize the objects, so we can deserialize them back, right?

vancexu: I agree the ideal case would be return `Map<String, Object>` to user, but because cadence client has no knowledge on the object type, it cannot deserialize bytes to correct object.

vancexu: Discussed with Liang offline and we have 2 proposals here with pros and cons. We'd like to here some input from you @mfateev .
This is about: get search attributes inside workflow. 
1. Liang prefer `Map<String, byte[]> getSearchAttributes();` instead of thrift `SearchAttributes` which contains `Map<String, ByteBuffer>`. Then user can use JsonDataConverter to convert `byte[]` to the concrete type. 
2. I was thinking just return thrift `SearchAttributes`, and provide util function `WorkflowUtils. getValueFromSearchAttributes()` for user to get concrete value. In this way, user keeps same approach when getting search attribute from List/Describe workflow APIs.

PR #420, opened 2019-11-24
Travis CI:  use 'https' instead of 'http'

sullis:


mfateev:
Do not check in until the build goes through.

Issue #398, opened 2019-10-01
Add context propagation support


mfateev:
Already added to Go client through https://github.com/uber-go/cadence-client/issues/729

izishared:
Great! We are waiting this function so much!

lan-n-tran:
Do we actually release this feature in Java? I think it will be really helpful.

longquanzheng:
@lan-n-tran yes it’s released

PR #428, opened 2020-01-02
Implement Context Propagation for Java

sdwr98:
This PR addresses #398 by providing context propagation support for the Java client.  

- Creates a ContextPropagation interface
- Uses the Header field to propagate context information from client -> workflow -> activity


src/main/java/com/uber/cadence/internal/replay/DecisionContext.java
```diff
@@ -80,6 +80,9 @@
    */
   SearchAttributes getSearchAttributes();
 
+  /** Propagates the context */
+  void propagateContext();
+
```
mfateev: This API should be more like getCurrentContext() as the DecisionContext API is asynchronous and shoudln't rely on thread locals.

src/main/java/com/uber/cadence/internal/sync/WorkflowRunnable.java
```diff
@@ -43,6 +43,7 @@ public WorkflowRunnable(
   @Override
   public void run() {
     try {
+
```
mfateev: ?

src/main/java/com/uber/cadence/worker/WorkerOptions.java
```diff
@@ -187,6 +190,12 @@ public Builder setEnableLoggingInReplay(boolean enableLoggingInReplay) {
       return this;
     }
 
+    /** Specifies the list of context propagators to use during this workflow. */
+    public Builder setContextPropagators(List<ContextPropagator> contextPropagators) {
+      this.contextPropagators = contextPropagators;
+      return this;
+    }
+
```
mfateev: I believe FactoryOptions is the better place for this.

src/main/java/com/uber/cadence/context/ContextPropagator.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.Map;
+
+public interface ContextPropagator {
+
+  /** Retrieves the current context for injection into the headers */
+  Map<String, byte[]> getCurrentContext();
+
+  /** Sets the context values with data from the headers */
+  void setCurrentContext(Map<String, byte[]> context);
+}
```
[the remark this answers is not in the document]
sdwr98: So are you suggesting something like transferContext(Thread) that the context propagator could use to manually transfer?  When would you envision this being used?  Local Activities?

src/main/java/com/uber/cadence/internal/sync/WorkflowRunnable.java
sdwr98: whoops, fixed that.

src/main/java/com/uber/cadence/worker/WorkerOptions.java
sdwr98: The reason I put them in the WorkerOptions is because then they're accessible when the StartWorkflowExecutionParameters get created (and thus I can set the Header).  If the ContextPropagators are in the FactoryOptions, what is your thought on how to access them during creation of the messages to the Cadence server?

src/main/java/com/uber/cadence/internal/sync/WorkflowThreadImpl.java
```diff
@@ -78,6 +78,10 @@ public void run() {
       MDC.put(LoggerTag.RUN_ID, decisionContext.getRunId());
       MDC.put(LoggerTag.TASK_LIST, decisionContext.getTaskList());
       MDC.put(LoggerTag.DOMAIN, decisionContext.getDomain());
+
+      // Process all the context propagators
+      decisionContext.propagateContext();
```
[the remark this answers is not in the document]
sdwr98: So you would like to see the context set explicitly as a property of the WorkflowThreadImpl at construction time (by the DeterministicRunner, maybe working our way up to SyncWorkflow)? And then invoke the ContextPropagator's methods to do whatever it is they do?

src/main/java/com/uber/cadence/activity/ActivityOptions.java
```diff
@@ -143,14 +149,21 @@ public Builder setRetryOptions(RetryOptions retryOptions) {
       return this;
     }
 
+    /** ContextPropagators help propagate the context from the workflow to the activities */
+    public Builder setContextPropagators(List<ContextPropagator> contextPropagators) {
+      this.contextPropagators = contextPropagators;
+      return this;
+    }
```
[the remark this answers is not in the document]
sdwr98: Done

src/main/java/com/uber/cadence/internal/sync/WorkflowThreadImpl.java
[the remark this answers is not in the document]
mfateev: Yes. The basic idea is that a thread should inherit context from its parent thread instead of always root context.

src/main/java/com/uber/cadence/context/ContextPropagator.java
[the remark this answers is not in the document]
mfateev: It is more like getCurentContext() returns opaque structure that can be used in setCurrentContext bypassing serialization/deserialization. And then have getCurrentContextAsMap, setCurrentContextFromMap to interface with headers. In Go it happens automatically when a context is created from the parent context.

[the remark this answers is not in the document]
sdwr98: Got it.  I did that with my latest change.

src/main/java/com/uber/cadence/internal/sync/WorkflowThreadImpl.java
```diff
@@ -143,6 +148,21 @@ public void setName(String name) {
         thread.setName(name);
       }
     }
+
+    public void propagateContext() {
+      // Loop through the context propagators and propagate
+      Map<String, Object> propagatedContexts = decisionContext.getPropagatedContexts();
```
mfateev: I might be confused. But it looks to me  that the context are coming not from a parent thread but from a single decision context shared by all threads. Am I missing something?

sdwr98: This might be a long answer but I hope it explains what my overall design was for this feature. 

This is my understanding (very high-level) of the flow of a workflow/activity execution: 

```
1 Workflow client ->
2    [Cadence workflow stub -> Cadence Server -> Workflow worker] ->
3       Workflow Implementation -> 
4            [Cadence activity stub -> Cadence Server -> Activity worker] ->
5                Activity Implementation
```
Where lines 1, 3, and 5 are code written by workflow/activity implementers, and lines 3 & 5 are Cadence infrastructure code.  

My thought was that context should flow only from line 1 to 3 to 5, and that the Cadence infrastructure/stubs/server should be "invisible" to the context propagation.  So when a workflow worker receives a `WorkflowExecutionStartedEvent`, I want to propagate the context into the thread that is executing the Workflow Implementation, not necessarily any of the intermediate threads spawned by the Cadence infrastructure. 

So I guess a) does that make sense, b) is that what you want, and c) am I missing something in that implementation?

mfateev: My idea (may be flawed) is that any thread should be able to access and modify the context. For example it can create a new tracing span and all the code that executes in that span including child threads should inherit from that context. So from the context propagation point of view the context should come from the parent thread, not from the workflowExecutionStartedEvent which injects context only into the main workflow thread.

sdwr98: I thought of doing that briefly, but that seemed to imply that everywhere Cadence started a new thread, we'd have to pass around the ContextPropagators and the context data.  That seemed like a lot of parameters moving around everywhere. 

I did have a thought this morning - I suppose I could use a ThreadLocal to hold the list of ContextPropagators and the context data.  But that would still require every Runnable object to take extra parameters for the context information.  

If you'd prefer that approach I can work on it.

sdwr98: Ahh - one thing I hadn't seen until now was the `Workflow.newPromise` and similar methods that allowed the Workflow to spawn off threads.  I'll do some more updates that take those into account.

mfateev: I don't think newPromise spawns threads. Only Async.... methods do. Use WorkflowThreadLocal instead of ThreadLocal.

sdwr98: Ok - I have it working now where any child threads will inherit the context from the parent threads.  Let me know what you think

build.gradle
```diff
@@ -37,7 +37,7 @@ googleJavaFormat {
 }
 
 group = 'com.uber.cadence'
-version = '2.6.3'
+version = '2.6.3-context-SNAPSHOT'
```
mfateev: probably need to revert

src/main/java/com/uber/cadence/context/ContextThreadLocal.java
```diff
@@ -0,0 +1,69 @@
...
+
+import com.uber.cadence.workflow.WorkflowThreadLocal;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Supplier;
+
+/** This class holds the current set of context propagators */
+public class ContextThreadLocal {
```
mfateev: Why is this public and not part of internal?

src/main/java/com/uber/cadence/worker/WorkerOptions.java
mfateev: Factory.newWorker can pass any parameters to a worker.

src/test/java/com/uber/cadence/internal/testing/WorkflowTestingTest.java
```diff
@@ -729,4 +735,116 @@ public void testMockedChildSimulatedTimeout() {
...
+    MDC.put("test", "testing123");
+    WorkflowClient client = testEnvironment.newWorkflowClient();
+    WorkflowOptions options =
+        new WorkflowOptions.Builder()
+            .setContextPropagators(Collections.singletonList(new TestContextPropagator()))
+            .build();
+    ParentWorkflow workflow = client.newWorkflowStub(ParentWorkflow.class, options);
+    String result = workflow.workflow("input1");
+    assertEquals("testing123testing123", result);
+  }
```
mfateev: Add test to propagating down to an activity as well as test that tests that a context value set in a thread is propagated to a child thread.

src/main/java/com/uber/cadence/context/ContextThreadLocal.java
sdwr98: No good reason.  Moved.

src/main/java/com/uber/cadence/worker/WorkerOptions.java
sdwr98: So if I use the flow FactoryOptions->Worker->SingleWorkerOptions, are you OK with that?

sdwr98:
OK - all comments addressed.  

What do I need to do for documentation purposes?  Javadoc? Anything for the cadenceworkflow.io site?

sdwr98:
Ok - Javadoc added.

mfateev:
I believe setting context propagator on activity and child workflow options is not required if it is set on the worker. Is it correct? Could you make it explicit in the JavaDoc as it looks like it is required.

sdwr98:
So  - setting the ContextPropagators on the FactoryOptions takes care of propagating the context *into* WorkflowWorkers, ActivityWorkers, and ChildWorkflowWorkers.  But propagating the context onward out of the WorkflowWorker currently requires explicit setting of the ContextPropagators on the ActivityOptions or ChildWorkflowOptions. 

Would you expect that when you set up an ActivityStub or ChildWorkflowStub inside a Workflow that has been configured with a ContextPropagator that by default they would inherit the same ContextPropagators?

mfateev:
Yes, exactly. We already do this for DataConverter for example. You can set it in the ActivityOptions and ChildWorkflowOptions, but by default it uses the one set on the worker.

So the only two places where setting propagators should be required is Worker.Factory and WorkflowClient.

sdwr98:
Do you want the ability to override context propagators on the ActivityOptions and ChildWorkflowOptions?  Because it looks like you can't do that for DataConverters...

sdwr98:
That might make more sense, now that I think about it - to treat the context propagators similarly to the data converters...

mfateev:
I think DataConverter should be overridable through Activity and Child workflow options. It looks like it is not, but I think we can add it in the future is such need arises. 

As far as context propagators let's allow overriding them, but by default use the one from worker.

sdwr98:
OK - code is implemented to use the ContextPropagator from the FactoryOptions by default, the documentation is updated to reflect that, and I've added 2 more test cases to test that scenario.  Thanks for pointing that out.

Issue #435, opened 2020-01-16
Automatically update IDL submodule on build 


RemcoBuddelmeijer:
Now that we have the IDL submodule (#427) in the Java project root folder it should also automatically update this submodule when building.
My suggestion is to do this in the `gradle.build` file.
This will not only make it easier for anyone to just code and to not have to worry about having to initialize and update the submodule him/herself.

RemcoBuddelmeijer:
In order to achieve the purpose to automatically sync, the master branch should only be used when the master branch of the Java client has been updated to support the new thrift files

PR #427, opened 2019-12-31
Use cadence-idl repo IDL storage

wxing1292:
* Delete thrift dir

PR #436, opened 2020-01-16
Automatically update idl submodule on build #435

RemcoBuddelmeijer:
Now that we have the IDL submodule (#427) in the Java project root folder it should also automatically update this submodule when building.
My suggestion is to do this in the `gradle.build` file.
This will not only make it easier for anyone to just code and to not have to worry about having to initialize and update the submodule him/herself.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ target
 /build
 /out
 dummy
+src/main/idls/*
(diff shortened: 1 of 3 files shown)


RemcoBuddelmeijer:
Thank you @mfateev 
Closing PR

PR #453, opened 2020-02-27
Get raw history integration (Testing not ready for review)

luowenking1:


src/main/java/com/uber/cadence/internal/testservice/TestWorkflowService.java
```diff
@@ -328,6 +328,28 @@ public GetWorkflowExecutionHistoryResponse GetWorkflowExecutionHistory(
...
 
+  @Override
+  public PollForWorkflowExecutionRawHistoryResponse PollForWorkflowExecutionRawHistory(
+      PollForWorkflowExecutionRawHistoryRequest getRequest)
+      throws BadRequestError, InternalServiceError, EntityNotExistsError, ServiceBusyError,
+          TException {
+    ExecutionId executionId = new ExecutionId(getRequest.getDomain(), getRequest.getExecution());
+    TestWorkflowMutableState mutableState = getMutableState(executionId);
+
+    return null;
```
mfateev: Please never do this. We already had customers spending time figuring out why service API silently ignores a request. Always throw UnsupportedOperationException("not implemented") instead. 

Also I would just implement these methods. It would take 15 minutes to copy/paste the code from other methods and replace names.

PR #458, opened 2020-03-04
Add activity type and workflow type tags for activity e2e metrics

meiliang86:


src/main/java/com/uber/cadence/internal/worker/ActivityWorker.java
```diff
@@ -155,43 +150,52 @@ private TaskHandlerImpl(ActivityTaskHandler handler) {
...
+    public void handle(PollForActivityTaskResponse task) throws Exception {
       Scope metricsScope =
           options
               .getMetricsScope()
               .tagged(
-                  ImmutableMap.of(MetricsTag.ACTIVITY_TYPE, task.task.getActivityType().getName()));
+                  ImmutableMap.of(
+                      MetricsTag.ACTIVITY_TYPE,
+                      task.getActivityType().getName(),
+                      MetricsTag.WORKER_TYPE,
```
vancexu: Should this be `WorkflowType` or `WORKER_TYPE`?

src/main/java/com/uber/cadence/internal/worker/ActivityWorker.java
```diff
@@ -155,43 +150,52 @@ private TaskHandlerImpl(ActivityTaskHandler handler) {
...
         sw.stop();
 
         sw = metricsScope.timer(MetricsType.ACTIVITY_RESP_LATENCY).start();
-        sendReply(task.task, response, metricsScope);
+        sendReply(task, response, metricsScope);
         sw.stop();
 
-        task.markDone();
+        metricsScope
+            .timer(MetricsType.ACTIVITY_E2E_LATENCY)
```
vancexu: Looks like `ACTIVITY_E2E_LATENCY` is already recoded in activity poll method. 
The other concern on `System.nanoTime() - task.getScheduledTimestampOfThisAttempt())` might not be accurate considering time skew between server and worker

meiliang86: It's moved from there. Reason is that this is more accurate (e2e time should start from schedule time). Yes I know about time skew, but that's the best we can do.

src/main/java/com/uber/cadence/internal/worker/ActivityWorker.java
meiliang86: Yes it should be workflow type. Fixed.

src/main/java/com/uber/cadence/internal/worker/ActivityWorker.java
```diff
@@ -155,43 +150,52 @@ private TaskHandlerImpl(ActivityTaskHandler handler) {
...
-      propagateContext(task.task);
+      propagateContext(task);
 
       try {
         Stopwatch sw = metricsScope.timer(MetricsType.ACTIVITY_EXEC_LATENCY).start();
-        ActivityTaskHandler.Result response = handler.handle(task.task, metricsScope, false);
+        ActivityTaskHandler.Result response = handler.handle(task, metricsScope, false);
         sw.stop();
 
         sw = metricsScope.timer(MetricsType.ACTIVITY_RESP_LATENCY).start();
```
dhiren-singh: Manual activity completion isn't captured by these metrics.

src/main/java/com/uber/cadence/internal/worker/ActivityWorker.java
```diff
@@ -155,43 +150,52 @@ private TaskHandlerImpl(ActivityTaskHandler handler) {
...
+      MDC.put(LoggerTag.ACTIVITY_ID, task.getActivityId());
+      MDC.put(LoggerTag.ACTIVITY_TYPE, task.getActivityType().getName());
+      MDC.put(LoggerTag.WORKFLOW_ID, task.getWorkflowExecution().getWorkflowId());
+      MDC.put(LoggerTag.RUN_ID, task.getWorkflowExecution().getRunId());
 
-      propagateContext(task.task);
+      propagateContext(task);
 
       try {
         Stopwatch sw = metricsScope.timer(MetricsType.ACTIVITY_EXEC_LATENCY).start();
```
dhiren-singh: Manual activity completion isn't captured by these metrics.

Issue #442, opened 2020-02-11
Expose parent workflow info (ID and RunID) through GetWorkflowInfo API 


meiliang86:
Similar to what we did in Go client:
https://github.com/uber-go/cadence-client/issues/635

PR #462, opened 2020-03-05
Add parent workflow ID and run ID to WorkflowInfo (#442)

emrahs:
Prior to this change, child workflows didn't have a way to retrieve the workflow ID and
run ID of the parent workflow. This change adds getters to the WorkflowInfo interface
to expose the information that exists in the WorkflowExecutionStartedEvent.

Resolves: #442

### DIFF
--- a/src/main/java/com/uber/cadence/internal/sync/WorkflowInfoImpl.java
+++ b/src/main/java/com/uber/cadence/internal/sync/WorkflowInfoImpl.java
@@ -69,12 +69,12 @@ final class WorkflowInfoImpl implements WorkflowInfo {
   @Override
   public String getParentWorkflowId() {
     WorkflowExecution parentWorkflowExecution = context.getParentWorkflowExecution();
-    return parentWorkflowExecution == null ? null : parentWorkflowExecution.workflowId;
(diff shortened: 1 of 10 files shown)


src/main/java/com/uber/cadence/internal/sync/WorkflowInfoImpl.java
```diff
@@ -64,4 +65,16 @@ public Duration getExecutionStartToCloseTimeout() {
...
+  @Override
+  public String getParentWorkflowId() {
+    WorkflowExecution parentWorkflowExecution = context.getParentWorkflowExecution();
+    return parentWorkflowExecution == null ? null : parentWorkflowExecution.workflowId;
+  }
+
+  @Override
+  public String getParentRunId() {
+    WorkflowExecution parentWorkflowExecution = context.getParentWorkflowExecution();
+    return parentWorkflowExecution == null ? null : parentWorkflowExecution.workflowId;
```
vancexu: should be runId

emrahs: good catch!

PR #470, opened 2020-03-09
Integration get raw history

luowenking1:


src/main/java/com/uber/cadence/serviceclient/WorkflowServiceTChannel.java
```diff
@@ -827,31 +832,120 @@ private StartWorkflowExecutionResponse startWorkflowExecution(
   @Override
   public GetWorkflowExecutionHistoryResponse GetWorkflowExecutionHistory(
       GetWorkflowExecutionHistoryRequest request) throws TException {
+
```
venkat1109: What is the rollback plan in the case of following scenarios :

- We rollback server to a version that no longer supports this API and client breaks
- Customer picks up new version of client library, runs into performance issues
- Customer picks up new version of client library and we discover an issue with RawHistory 

I strongly recommend adding a  feature flag to turn this on/off in addition to checking for version compatibility with the server.

Issue #443, opened 2020-02-14
Integrate With New Get History Raw API


luowenking1:
Integrate with the new get raw history API, since currently on the server side the history events are deserialized when loaded from DB and serialized again when dispatching to client.

Server side corresponding issue: uber/cadence#2644

PR #472, opened 2020-03-12
Integration get raw history #443

luowenking1:


src/main/java/com/uber/cadence/internal/common/InternalUtils.java
```diff
@@ -18,19 +18,18 @@
 package com.uber.cadence.internal.common;
 
 import com.google.common.base.Defaults;
-import com.uber.cadence.SearchAttributes;
-import com.uber.cadence.TaskList;
-import com.uber.cadence.TaskListKind;
+import com.uber.cadence.*;
```
meiliang86: Please do not collapse the imports.

meiliang86: How do you test the new logic?

meiliang86:
How do you test the new logic?

luowenking1:
> How do you test the new logic?

Earlier the server change will always return the raw data, so I added the tests to mock the raw results and they are working fine. I removed them when we decides the change because of compatibility of server and client. However, I just revised the existing one for raw data deserialization testing which is working fine. For integration testing,  now the returned values from server depends on the value of a new flag SendRawHistoryData on server side. Seems we need to set up some different dockers for server and use them for testing, but not sure how to proceed with that yet. BTW, the pr on server was merged yesterday.

PR #473, opened 2020-03-23
Retry getting result from passive in case of replication delay

meiliang86:
Cadence supports auto forwarding for startWorkflow execution, but not for getHistory. This poses an issue if user want to start workflow and block on getting the result from the passive side, as the getHistory call will likely to fail due to the replication lag. This also can cause additional load to archival system for archive enabled domain, as cadence will fallback to query archival if it cannot find it in its primary persistence.

We are doing the following changes to mitigate this issue:
1. In GetWorkflowExecutionHistory request, add a flag to indicate skip querying archival.
2. If history does not exist and domain is not active in the region that serves the request, return this information in the error.
3. Client continue to retry and long poll in case of this error.

### DIFF
--- a/src/main/java/com/uber/cadence/internal/common/InternalUtils.java
+++ b/src/main/java/com/uber/cadence/internal/common/InternalUtils.java
@@ -19,23 +19,23 @@ package com.uber.cadence.internal.common;
 
 import com.google.common.base.Defaults;
 import com.uber.cadence.DataBlob;
+import com.uber.cadence.EntityNotExistsError;
 import com.uber.cadence.History;
 import com.uber.cadence.HistoryEvent;
 import com.uber.cadence.HistoryEventFilterType;
 import com.uber.cadence.SearchAttributes;
 import com.uber.cadence.TaskList;
 import com.uber.cadence.TaskListKind;
-import com.uber.cadence.EntityNotExistsError;
 import com.uber.cadence.converter.DataConverter;
 import com.uber.cadence.converter.JsonDataConverter;
 import com.uber.cadence.internal.worker.Shutdownable;
 import com.uber.cadence.workflow.WorkflowMethod;
 import java.lang.reflect.Method;
 import java.nio.ByteBuffer;
-import java.util.Arrays;
 import java.util.ArrayList;
-import java.util.List;
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
@@ -193,7 +193,8 @@ public final class InternalUtils {
       try {
         blob.setData(serializer.serialize(event));
       } catch (org.apache.thrift.TException err) {
-        throw new EntityNotExistsError("Deserialize blob data to history event failed with unknown error");
+        throw new EntityNotExistsError(
+            "Deserialize blob data to history event failed with unknown error");
       }
       blobs.add(blob);
     }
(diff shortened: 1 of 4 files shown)


src/main/java/com/uber/cadence/internal/common/InternalUtils.java
```diff
@@ -193,7 +193,8 @@ public static History DeserializeFromBlobToHistoryEvents(
       try {
         blob.setData(serializer.serialize(event));
       } catch (org.apache.thrift.TException err) {
-        throw new EntityNotExistsError("Deserialize blob data to history event failed with unknown error");
+        throw new EntityNotExistsError(
+            "Deserialize blob data to history event failed with unknown error");
```
emrahs: While you are touching this line, consider appending err.getMessage() to this error message for visibility.

src/main/java/com/uber/cadence/internal/common/WorkflowExecutionUtils.java
```diff
@@ -191,31 +194,46 @@ public static HistoryEvent getInstanceCloseEvent(
...
+                + unit);
+      }
+
       GetWorkflowExecutionHistoryRequest r = new GetWorkflowExecutionHistoryRequest();
       r.setDomain(domain);
       r.setExecution(workflowExecution);
       r.setHistoryEventFilterType(HistoryEventFilterType.CLOSE_EVENT);
       r.setNextPageToken(pageToken);
       r.setWaitForNewEvent(true);
+      r.setSkipArchival(true);
```
vancexu: When will `SkipArchival` be true? Is SkipArchival always true for long-poll?

src/main/java/com/uber/cadence/internal/common/WorkflowExecutionUtils.java
```diff
@@ -191,31 +194,46 @@ public static HistoryEvent getInstanceCloseEvent(
...
+            && !e.activeCluster.equals(e.currentCluster)) {
+          // Current cluster is passive cluster. Execution might not exist because of replication
+          // lag. Wait for a little bit and retry.
+          try {
+            Thread.sleep(ENTITY_NOT_EXIST_RETRY_WAIT_MILLIS);
+          } catch (InterruptedException ie) {
+            // Throw entity not exist here.
+            throw e;
+          }
+          continue;
```
emrahs: With your change, we no longer throw an `EntityNotExistsError` when the passive cluster is queried; except the `InterruptedException` case above. In all other cases where we simply `continue` and retry, we would eventually throw a `TimeoutException`. I find that a bit misleading.

I'd consider throwing the most recent `EntityNotExistsError` instead of a new `TimeoutException` when we give up on retries due to timeout. Is there any reason for not doing that?

src/main/java/com/uber/cadence/internal/common/WorkflowExecutionUtils.java
meiliang86: When user want to get archival history from CLI/UI. This method is shared by user and worker.

src/main/java/com/uber/cadence/internal/common/WorkflowExecutionUtils.java
meiliang86: Done.

src/main/java/com/uber/cadence/internal/common/InternalUtils.java
meiliang86: This error type is wrong. I've fixed it.

meiliang86:
Buildkite passed https://buildkite.com/uberopensource/cadence-java-client/builds/347

PR #474, opened 2020-03-24
Fix race in local activity completion

meiliang86:


src/main/java/com/uber/cadence/internal/replay/ReplayDecider.java
```diff
@@ -100,6 +104,19 @@
...
+          lock.lock();
+          try {
+            processEvent(historyEvent);
+          } finally {
+            lock.unlock();
+          }
+        };
+  }
+
+  Lock getLock() {
```
vancexu: nit: remove if not used.

meiliang86: It's used in clock decision context.

PR #478, opened 2020-04-02
Opentracing

sdwr98:
Implements OpenTracing support using the ContextPropagator functionality recently released.


src/main/java/com/uber/cadence/internal/sync/WorkflowThreadImpl.java
```diff
@@ -92,13 +99,26 @@ public void run() {
       ContextThreadLocal.setContextPropagators(this.contextPropagators);
       ContextThreadLocal.propagateContextToCurrentThread(this.propagatedContexts);
 
-      try {
+      // Set up an opentracing span
```
mfateev, line 102: Changing the code here for each new propagator is not going to scale. Can we do it in the "propagateContexttoTheCurrentThread"? Otherwise we need to extend the propagator API to avoid such "workarounds".

src/main/java/com/uber/cadence/internal/worker/ActivityWorker.java
```diff
@@ -175,7 +181,18 @@ public void handle(PollForActivityTaskResponse task) throws Exception {
 
       propagateContext(task);
 
-      try {
+      // Set up an opentracing span
```
mfateev, line 184: ditto. We don't want to change any of the client code to support a specific context propagator.

mfateev (changes requested):

src/main/java/com/uber/cadence/internal/sync/WorkflowThreadImpl.java
sdwr98: Yeah, I wasn't super happy about having to do this.  Unlike Go, there's nowhere to drop the span information outside the context of an active span.  

One option I suppose would be to implement some sort of `ContextPropagator.cleanup()` method that got invoked at the end of a Workflow/Activity - in that case we could start/end the span solely inside the propagator. 

Thoughts?

sdwr98: @mfateev OK - I went with option 1, and now there's no OpenTracing-specific code in the core library except where we automatically add an OpenTracingContextPropagator if one doesn't exist.  Would be curious to get your take on that decision

shijiesheng:

PR #493, opened 2020-06-01
Fix bug in releasing buffer for Async GetWorkflowExecutionHistory

meiliang86:


src/main/java/com/uber/cadence/client/WorkflowOptions.java
```diff
@@ -124,9 +124,10 @@ public Builder setWorkflowId(String workflowId) {
      * conditions Cadence allows two workflows with the same domain and workflow id run
      * simultaneously.
      * <li>
-     * <ul>
-     *   AllowDuplicateFailedOnly is a default value. It means that workflow can start if
-     *   previous run failed or was cancelled or terminated.
+     *
+     *     <ul>
```
vancexu, line 128: nit: indent?

meiliang86: This is somehow changed automatically by the code formatter through gradlew build.

meiliang86:
> Will other API have similar issues?

No. We only implemented one Async API which is this one.

src/main/java/com/uber/cadence/serviceclient/WorkflowServiceTChannel.java
```diff
@@ -2336,18 +2336,17 @@ public void GetWorkflowExecutionHistory(
...
-                return;
-              })
-          .exceptionally(
-              (e) -> {
-                log.error("Unexpected error in GetWorkflowExecutionHistory", e);
-                return null;
-              });
-    } finally {
-      if (response != null && response.isDone()) {
-        response.join().release();
```
mkolodezny, line 2348: nit: so the response is never released if not done immediately here?

meiliang86: That's the bug to be fixed in this diff, right?

meiliang86:
Buildkite passed: https://buildkite.com/uberopensource/cadence-java-client/builds/374

PR #501, opened 2020-06-08
Honor user timeout for get workflow result

vancexu:


meiliang86:
Change the sync method as well? Search for the reference for retryParameters

vancexu:
closed in favor of https://github.com/uber/cadence-java-client/pull/504

PR #503, opened 2020-06-09
Add GetWorkflowExecutionHistoryWithTimeout

vancexu:


src/main/java/com/uber/cadence/serviceclient/WorkflowServiceTChannel.java
```diff
@@ -2297,15 +2297,46 @@ public void StartWorkflowExecution(
     throw new UnsupportedOperationException("not implemented");
   }
 
+  private Long validateAndUpdateTimeout(Long timeoutInMillis) {
+    if (timeoutInMillis <= 0 || timeoutInMillis == Long.MAX_VALUE) {
```
meiliang86, line 2301: if (timeoutInMillis <= 0 || timeoutInMillis > options.getRpcLongPollTimeoutMillis()) {

let's cap it at options.getRpcLongPollTimeoutMillis()

vancexu: I feel cap at cadence default settings may surprise users. Since we add the timeout explicitly in API, would it make more sense to prioritize use input?

vancexu:
closed in favor of https://github.com/uber/cadence-java-client/pull/504

Issue #498, opened 2020-06-07
Honor timeout while getting workflow result


meiliang86:
The retryOption we use for GetWorkflowExecutionHistory and getWorkflowExecutionHistoryAsync has infinite retry window. We should set expiration correctly based on the timeout value passed in by the user.

Also, 120s timeout seems to be too long for getResult. We should let user set it based on the timeout from getResult.

PR #504, opened 2020-06-09
Honor user timeout for get workflow result

vancexu:
https://github.com/uber/cadence-java-client/issues/498

### DIFF
--- a/src/main/java/com/uber/cadence/internal/common/WorkflowExecutionUtils.java
+++ b/src/main/java/com/uber/cadence/internal/common/WorkflowExecutionUtils.java
@@ -292,7 +292,7 @@ public class WorkflowExecutionUtils {
     request.setWaitForNewEvent(true);
     request.setNextPageToken(pageToken);
     CompletableFuture<GetWorkflowExecutionHistoryResponse> response =
-        getWorkflowExecutionHistoryAsync(service, request);
+        getWorkflowExecutionHistoryAsync(service, request, timeout, unit);
     return response.thenComposeAsync(
         (r) -> {
           long elapsedTime = System.currentTimeMillis() - start;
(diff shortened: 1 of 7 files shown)


src/main/java/com/uber/cadence/internal/common/WorkflowExecutionUtils.java
```diff
@@ -213,9 +213,12 @@ private static HistoryEvent getInstanceCloseEvent(
       r.setNextPageToken(pageToken);
       r.setWaitForNewEvent(true);
       r.setSkipArchival(true);
+      RetryOptions retryOptions = getRetryOptionTimeout(timeout, unit);
```
meiliang86, line 216: getRetryOptionTimeout -> getRetryOption

src/test/java/com/uber/cadence/internal/common/RetryerTest.java
```diff
@@ -105,4 +106,29 @@ public void testInterruptedException() throws InterruptedException {
...
+              () -> {
+                throw new IllegalArgumentException("simulated");
+              })
+          .get();
+      fail("unreachable");
+    } catch (ExecutionException e) {
+      assertTrue(e.getCause() instanceof IllegalArgumentException);
+      assertEquals("simulated", e.getCause().getMessage());
+    }
+    assertTrue(System.currentTimeMillis() - start < 500);
```
meiliang86, line 132: Instead of assert on time, increment a counter and assert on attempt?

vancexu: that is ideal but may need to update `retryWithResultAsync` signature to return attempts, since there is not way to have a counter in java closure. I will leave it for now.

vancexu:
buildkite retrigger

PR #506, opened 2020-06-17
Add async start to untype stub

vancexu:
Currently start workflow doesn't support async handler. In some cases, customer may be super sensitive to latency and want start to be async. 
This PR add an API for async start in untype stub. 

Next step in separate PRs:
~~add startAsyncWithTimeout~~ (also done in this PR per request)
add async startAndGetResult

### DIFF
--- a/src/main/java/com/uber/cadence/client/WorkflowStub.java
+++ b/src/main/java/com/uber/cadence/client/WorkflowStub.java
@@ -67,6 +67,8 @@ public interface WorkflowStub {
 
   WorkflowExecution start(Object... args);
 
+  CompletableFuture<WorkflowExecution> startAsync(Object... args);
+
   WorkflowExecution signalWithStart(String signalName, Object[] signalArgs, Object[] startArgs);
 
   Optional<String> getWorkflowType();
--- a/src/main/java/com/uber/cadence/client/WorkflowStub.java
+++ b/src/main/java/com/uber/cadence/client/WorkflowStub.java
@@ -69,6 +69,8 @@ public interface WorkflowStub {
 
   CompletableFuture<WorkflowExecution> startAsync(Object... args);
 
+  CompletableFuture<WorkflowExecution> startAsyncWithTimeout(Long timeoutInMillis, Object... args);
+
   WorkflowExecution signalWithStart(String signalName, Object[] signalArgs, Object[] startArgs);
 
   Optional<String> getWorkflowType();
--- a/src/main/java/com/uber/cadence/client/WorkflowStub.java
+++ b/src/main/java/com/uber/cadence/client/WorkflowStub.java
@@ -69,7 +69,8 @@ public interface WorkflowStub {
 
   CompletableFuture<WorkflowExecution> startAsync(Object... args);
 
-  CompletableFuture<WorkflowExecution> startAsyncWithTimeout(Long timeoutInMillis, Object... args);
+  CompletableFuture<WorkflowExecution> startAsyncWithTimeout(
+      long timeout, TimeUnit unit, Object... args);
 
   WorkflowExecution signalWithStart(String signalName, Object[] signalArgs, Object[] startArgs);
 
--- a/src/main/java/com/uber/cadence/serviceclient/WorkflowServiceTChannel.java
+++ b/src/main/java/com/uber/cadence/serviceclient/WorkflowServiceTChannel.java
@@ -2301,12 +2301,29 @@ public class WorkflowServiceTChannel implements IWorkflowService {
   @Override
   public void StartWorkflowExecution(
       StartWorkflowExecutionRequest startRequest, AsyncMethodCallback resultHandler) {
+    startWorkflowExecution(startRequest, resultHandler, null);
+  }
+
+  @Override
+  public void StartWorkflowExecutionWithTimeout(
+      StartWorkflowExecutionRequest startRequest,
+      AsyncMethodCallback resultHandler,
+      Long timeoutInMillis) {
+    startWorkflowExecution(startRequest, resultHandler, timeoutInMillis);
+  }
+
+  public void startWorkflowExecution(
+      StartWorkflowExecutionRequest startRequest,
+      AsyncMethodCallback resultHandler,
+      Long timeoutInMillis) {
 
     startRequest.setRequestId(UUID.randomUUID().toString());
+    timeoutInMillis = validateAndUpdateTimeout(timeoutInMillis, options.getRpcTimeoutMillis());
     ThriftRequest<WorkflowService.StartWorkflowExecution_args> request =
         buildThriftRequest(
             "StartWorkflowExecution",
-            new WorkflowService.StartWorkflowExecution_args(startRequest));
+            new WorkflowService.StartWorkflowExecution_args(startRequest),
+            timeoutInMillis);
 
     CompletableFuture<ThriftResponse<WorkflowService.StartWorkflowExecution_result>> response =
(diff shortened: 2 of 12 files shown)


src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -951,6 +951,34 @@ public void testSyncUntypedAndStackTrace() {
...
+
+    try {
+      WorkflowExecution execution = future.get();
+      WorkflowStub workflowStub =
+          workflowClient.newUntypedWorkflowStub(execution, Optional.empty());
+      String stackTrace = workflowStub.query(WorkflowClient.QUERY_TYPE_STACK_TRACE, String.class);
+      assertTrue(stackTrace, stackTrace.contains("WorkflowTest$TestSyncWorkflowImpl.execute"));
+      assertTrue(stackTrace, stackTrace.contains("activityWithDelay"));
+      // Test stub created from workflow execution.
+      workflowStub =
```
meiliang86, line 970: This is not needed.

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -951,6 +951,34 @@ public void testSyncUntypedAndStackTrace() {
...
+  public void testUntypedAsyncStart() {
+    startWorkerFor(TestSyncWorkflowImpl.class);
+    WorkflowStub startWorkflowStub =
+        workflowClient.newUntypedWorkflowStub(
+            "TestWorkflow1::execute", newWorkflowOptionsBuilder(taskList).build());
+    CompletableFuture<WorkflowExecution> future = startWorkflowStub.startAsync(taskList);
+
+    try {
+      WorkflowExecution execution = future.get();
+      WorkflowStub workflowStub =
```
meiliang86, line 964: Set workflowExecution on completion in WorkflowStubImpl.startAsyncWithOptions, so we can reuse the existing stub.

vancexu: make sense. Done

vancexu:
> Let's add timeout for start API in this PR, as we should share the same code path almost everywhere.
> The one without timeout is basically calling the one with timeout and set timeout = Long.MAX.

Sure, added.

src/main/java/com/uber/cadence/client/WorkflowStub.java
```diff
@@ -67,6 +67,10 @@
 
   WorkflowExecution start(Object... args);
 
+  CompletableFuture<WorkflowExecution> startAsync(Object... args);
```
meiliang86, line 70: This seems to be only used by test. Remove?

src/main/java/com/uber/cadence/client/WorkflowStub.java
```diff
@@ -67,6 +67,10 @@
...
+
+  CompletableFuture<WorkflowExecution> startAsyncWithTimeout(Long timeoutInMillis, Object... args);
```
meiliang86, line 72: If you remove the method above, you can just call this startAsync

src/main/java/com/uber/cadence/serviceclient/WorkflowServiceTChannel.java
```diff
@@ -2300,9 +2300,84 @@ private ListTaskListPartitionsResponse listTaskListPartitions(
...
+
+  @Override
+  public void StartWorkflowExecutionWithTimeout(
+      StartWorkflowExecutionRequest startRequest,
+      AsyncMethodCallback resultHandler,
+      Long timeoutInMillis) {
+    startWorkflowExecution(startRequest, resultHandler, timeoutInMillis);
+  }
+
+  public void startWorkflowExecution(
```
meiliang86, line 2315: This can be private.

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -938,6 +938,34 @@ public void testSyncUntypedAndStackTrace() {
...
+    startWorkerFor(TestSyncWorkflowImpl.class);
+    WorkflowStub workflowStub =
+        workflowClient.newUntypedWorkflowStub(
+            "TestWorkflow1::execute", newWorkflowOptionsBuilder(taskList).build());
+    Long timeout = new Long(200);
+    CompletableFuture<WorkflowExecution> future =
+        workflowStub.startAsyncWithTimeout(timeout, taskList);
+    testUntypedAndStackTraceHelper(workflowStub, future.get());
+  }
+
```
meiliang86, line 965: Can you add a negative test where start returns exception? For example try to start workflow twice and expect the second call to error out.

meiliang86:
Looks good now. Some minor comments.

src/main/java/com/uber/cadence/client/WorkflowStub.java
vancexu: It would be good to allow user just start without timeout, so will keep both APIs

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
vancexu: Added. 
Also added another test to first start then getResult both in async manners.

vancexu:
build pass here https://buildkite.com/uberopensource/cadence-java-client/builds/400#038ceef3-cfd2-4660-9068-8f3915cf96dc. Not sure why github build status doesn't reflect it.

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -938,6 +939,78 @@ public void testSyncUntypedAndStackTrace() {
...
+    CompletableFuture<WorkflowExecution> future = newWorkflowStub.startAsync(taskList);
+    try {
+      future.get();
+    } catch (ExecutionException e) {
+      assertTrue(e.getCause() instanceof WorkflowExecutionAlreadyStartedError);
+    }
+  }
+
+  @Test
+  public void testUntypedAsyncStartAndGetResult() throws InterruptedException {
```
meiliang86, line 996:
```
  @Test
  public void testUntypedAsyncStartAndGetResult() throws InterruptedException {
    startWorkerFor(TestSyncWorkflowImpl.class);
    String wfType = "TestWorkflow1::execute";
    long timeoutMillis = 5000;
    long startTime = System.currentTimeMillis();
    WorkflowStub startStub =
        workflowClient.newUntypedWorkflowStub(wfType, newWorkflowOptionsBuilder(taskList).build());
    CompletableFuture<String> resultFuture =
        startStub
            .startAsyncWithTimeout(timeoutMillis, TimeUnit.MILLISECONDS, taskList)
            .thenCompose(
                execution -> {
                  long remainingTimeMillis =
                      timeoutMillis - (System.currentTimeMillis() - startTime);
                  if (remainingTimeMillis <= 0) {
                    CompletableFuture<String> f = new CompletableFuture<>();
                    f.completeExceptionally(new TimeoutException());
                    return f;
                  }

                  WorkflowStub resultStub =
                      workflowClient.newUntypedWorkflowStub(execution, Optional.of(wfType));
                  return resultStub.getResultAsync(
                      remainingTimeMillis, TimeUnit.MILLISECONDS, String.class);
                });
    try {
      assertEquals("activity10", resultFuture.get());
    } catch (ExecutionException e) {
      fail("unreachable");
    }
  }
```

vancexu: build passed, ready for land

Issue #397, opened 2019-09-30
Metrics names are not compatible with prometheus


meiliang86:
Prometheus does not allow dash in metrics names.

szaluzhskiy:
Any forecast on resolution of this issue?

PR #507, opened 2020-06-18
Fix prometheus reporting issue

AleksandrKolosov:
MetricsType.ACTIVITY_SCHEDULED_TO_START_LATENCY is reported twice with a different set of tags.
It causes prometheus to fail with an IllegalArgumentException: "Prometheus requires that all meters with the same name have the same set of tag keys."
Reporter sets the fixed set of tags for each metric now.

Improved reporter to track gauges as well.


AleksandrKolosov:
Relates to https://github.com/uber/cadence-java-client/issues/397

PR #508, opened 2020-06-19
Promise.allOf should not block on empty input

### LINKED ISSUE #480: Bug in AllOfFuture
If all the Promises provided to `Promise.allOf` are already complete (e.g. all created using Workflow.newPromise(value)`) then the returned Promise will never complete.

This appears to be due to a gap in the logic in AllOfFuture: https://github.com/uber/cadence-java-client/blob/master/src/main/java/com/uber/cadence/internal/sync/AllOfFuture.java#L61

Here, the "already complete" code path sets the result in the array, but does not handle marking the internal CompletablePromise `impl` (https://github.com/uber/cadence-java-client/blob/master/src/main/java/com/uber/cadence/internal/sync/AllOfFuture.java#L33) as complete once all the results have been set.

Two options to fix this come to mind:
1) check the result array after all promises have been added - if they are all done, mark the `impl` as complete. If they are not all done, then there is at least one promise that will execute the `handle` block and mark the result as complete.
2) get rid of the `if (f.isCompleted())` branch altogether, and route all promises through the `handle` logic. https://github.com/uber/cadence-java-client/blob/master/src/main/java/com/uber/cadence/internal/sync/AllOfFuture.java#L60

The following workflow can be used to demonstrate the issue:

```java
public class PromiseFlow {
    public String execute(String input) {
        CompletablePromise<String> inputP = Workflow.newPromise();
        inputP.complete(input);
        var resultCar1 = inputP.thenApply(in -> in + in);
        var resultCar2 = inputP.thenApply(in -> in + in + in);
        var list = Promise.allOf(List.of(resultCar1, resultCar2)).get();
        return String.join(",", list);
    }
}
```

Compare to this to a version with a workaround using recursion:

```java
public class PromiseFlow {
  public String execute(String input) {
        CompletablePromise<String> inputP = Workflow.newPromise();
        inputP.complete(input);
        var resultCar1 = inputP.thenApply(in -> in + in);
        var resultCar2 = inputP.thenApply(in -> in + in + in);
        var list = allOfImpl(List.of(resultCar1, resultCar2)).get();
        return String.join(",", list);
    }
  
  private <T> Promise<List<T>> allOfImpl(List<Promise<T>> promises) {
        if (promises.isEmpty()) {
            return Workflow.newPromise(Collections.emptyList());
        }

        if (promises.size() == 1) {
            return promises.get(0).thenApply(result -> new ArrayList<T>(List.of(result)));
        }

        var remainder = promises.stream().skip(1).collect(toList());
        return allOfImpl(remainder).thenCompose(list -> promises.get(0).thenApply(result -> {
            list.add(result);
            return list;
        }));
    }
}
```

Thank you!

rlei:
Fixes #480: when `Promise.allOf` is called with an empty collection, the returned promise should not block forever, but just complete immediately.

### DIFF
--- a/src/main/java/com/uber/cadence/internal/sync/AllOfFuture.java
+++ b/src/main/java/com/uber/cadence/internal/sync/AllOfFuture.java
@@ -23,6 +23,7 @@ import com.uber.cadence.workflow.Promise;
 import com.uber.cadence.workflow.Workflow;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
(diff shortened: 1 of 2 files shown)


ilmn-aeolus:
I don't see how this addresses https://github.com/uber/cadence-java-client/issues/480.

https://github.com/uber/cadence-java-client/issues/480 is not about an empty list of promises, but about a list of promises that are already complete.

PR #511, opened 2020-06-30
Change activity discovery logic

WToma:
Update `addActivityImplementation` to only add methods as activities
which have the `ActivityMethod` annotation.

The current logic will register all interface-supplied methods as activities.
This becomes problematic when a Scala `trait` is extended multiple times
in the inheritance chain, and that `trait` supplies some methods to the class
(in my case, these would be logging utils).


vancexu:
Let us know what do you think. If nothing else, we may close this PR

WToma:
Ok, I guess I misunderstood what `@ActivityMethod` was for then. Just curious, what's the expected usage of that annotation then?

vancexu:
> Ok, I guess I misunderstood what `@ActivityMethod` was for then. Just curious, what's the expected usage of that annotation then?

Used for overwrite activity options like type name. check more on https://www.javadoc.io/doc/com.uber.cadence/cadence-client/latest/index.html

PR #534, opened 2020-10-01
Fix get raw history

vancexu:
Currently GetRawHistory is not working by going to endless loop in `getInstanceCloseEventAsync` then timeout.
This PR fix this issue.
Manually tests using existing tests.
(Because on server side raw history is disabled by default, testing this feature requires turning on this feature and set `useDockerService` to true in WorkflowTest

### DIFF
--- a/src/main/java/com/uber/cadence/internal/common/WorkflowExecutionUtils.java
+++ b/src/main/java/com/uber/cadence/internal/common/WorkflowExecutionUtils.java
@@ -311,7 +311,19 @@ public class WorkflowExecutionUtils {
                         + ", unit="
                         + unit));
           }
-          History history = r.getHistory();
+          History history;
+          if (r.getRawHistory() != null) {
+            try {
+              history =
+                  InternalUtils.DeserializeFromBlobToHistoryEvents(
+                      r.getRawHistory(), request.getHistoryEventFilterType());
+            } catch (TException e) {
+              e.printStackTrace();
+              throw new RuntimeException("Unable to deserialize raw history");
+            }
+          } else {
+            history = r.getHistory();
+          }
           if (history == null || history.getEvents().size() == 0) {
             // Empty poll returned
             return getInstanceCloseEventAsync(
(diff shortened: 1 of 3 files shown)


src/main/java/com/uber/cadence/internal/common/WorkflowExecutionUtils.java
```diff
@@ -311,7 +311,19 @@ private static HistoryEvent getInstanceCloseEvent(
                         + ", unit="
                         + unit));
           }
-          History history = r.getHistory();
```
mkolodezny, line 314: Great catch Bowei, thanks!
I believe the other response.getHistory usages needs to be updated the same way: e.g. https://github.com/uber/cadence-java-client/blob/5b28096165ee4bb075bb1a9ac10064916b7b070a/src/main/java/com/uber/cadence/internal/worker/WorkflowWorker.java#L131
https://github.com/uber/cadence-java-client/blob/243e758a0093a5916ec08f73a02c580ac2af52b1/src/main/java/com/uber/cadence/internal/common/WorkflowExecutionUtils.java#L248
.....

vancexu: Thanks for bring it up! 
I double checked and found they actually don't need to be updated because https://github.com/uber/cadence-java-client/pull/472 already handle it for sync `getWorkflowExecutionHistory`. 
but it help me realize that `WorkflowExecutionUtils` is not a good place as `WorkflowServiceTChannel`, so I move my fix there.

mkolodezny (approved):
Looks good! Def need to be able to catch such things upon bench testing before the release.

meiliang86 (approved):

meiliang86:
Thanks for the fix @vancexu !

vancexu:
Thanks @meiliang86 for helping the buildkite test issues.

vancexu:
https://buildkite.com/uberopensource/cadence-java-client/builds/449#_

PR #537, opened 2020-10-06
Fix ACTIVITY_E2E_LATENCY metrics

meiliang86:
Merge fix: https://github.com/temporalio/sdk-java/pull/68


vytautas-karpavicius (approved):

vancexu (approved):

src/main/java/com/uber/cadence/internal/common/Retryer.java
```diff
@@ -1,8 +1,8 @@
 /*
+ *  Modifications Copyright (c) 2017-2020 Uber Technologies Inc.
```
vancexu, line 2: You may need to update the license check also

Issue #416, opened 2019-11-14
interface not visible from class loader


mohsalxedflix:
I used the example at https://github.com/uber/cadence-java-samples/blob/master/src/main/java/com/uber/cadence/samples/hello/HelloActivity.java to try implement it with a Spring boot application.

I'm getting this following error:
```
java.lang.IllegalArgumentException: interface com.kloudlearn.newsfeed.workflow.HelloActivity$GreetingWorkflow is not visible from class loader
	at java.lang.reflect.Proxy$ProxyClassFactory.apply(Proxy.java:581)
	at java.lang.reflect.Proxy$ProxyClassFactory.apply(Proxy.java:557)
	at java.lang.reflect.WeakCache$Factory.get(WeakCache.java:230)
	at java.lang.reflect.WeakCache.get(WeakCache.java:127)
	at java.lang.reflect.Proxy.getProxyClass0(Proxy.java:419)
	at java.lang.reflect.Proxy.newProxyInstance(Proxy.java:719)
	at com.uber.cadence.internal.sync.WorkflowClientInternal.newWorkflowStub(WorkflowClientInternal.java:162)
	at com.uber.cadence.internal.sync.WorkflowClientInternal.newWorkflowStub(WorkflowClientInternal.java:146)
	at com.kloudlearn.newsfeed.service.PostService.workflowTest(PostService.java:101)
	at com.kloudlearn.newsfeed.service.PostService$$FastClassBySpringCGLIB$$79945899.invoke(<generated>)
	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:204)
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.invokeJoinpoint(CglibAopProxy.java:746)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:163)
	at org.springframework.aop.aspectj.MethodInvocationProceedingJoinPoint.proceed(MethodInvocationProceedingJoinPoint.java:88)
	at com.kloudlearn.newsfeed.aop.logging.LoggingAspect.logAround(LoggingAspect.java:85)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #539, opened 2020-10-07
Switched newProxyInstance to the correct class

meiliang86:
Merge fix: https://github.com/temporalio/sdk-java/pull/71
Background context:
https://stackoverflow.com/questions/211176/interface-is-not-visible-from-classloader-when-using-a-proxy/13379769#13379769


meiliang86:
Fixes https://github.com/uber/cadence-java-client/issues/416

vytautas-karpavicius (approved):

PR #540, opened 2020-10-08
Remove disable activity and workflow worker options

meiliang86:
- Workers will always start, and poller will start only if types are registered.
- Simplified worker start and shutdown logic
- Fixed a bug in worker.isSuspended()

Ref: https://github.com/temporalio/sdk-java/pull/33/commits/4c1d9b02f6e73eb44d22a2e17b226611663e3373

### DIFF
--- a/src/main/java/com/uber/cadence/internal/worker/LocalActivityWorker.java
+++ b/src/main/java/com/uber/cadence/internal/worker/LocalActivityWorker.java
@@ -34,16 +34,14 @@ import java.time.Duration;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.concurrent.TimeUnit;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.LongSupplier;
 
-public final class LocalActivityWorker implements SuspendableWorker {
+public final class LocalActivityWorker extends SuspendableWorkerBase {
 
   private static final String POLL_THREAD_NAME_PREFIX = "Local Activity Poller taskList=";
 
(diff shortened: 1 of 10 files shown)


src/main/java/com/uber/cadence/internal/worker/LocalActivityWorker.java
```diff
@@ -85,46 +85,73 @@ public void start() {
 
   @Override
   public boolean isStarted() {
+    if (poller == null) {
```
vytautas-karpavicius, line 88: It seems like poller is `NoopSuspendableWorker` initially (that is when it is not started). We should not need those null checks here.

meiliang86: Good catch. Remove all of these and moved common methods implementations into an abstract base class.

src/main/java/com/uber/cadence/worker/WorkerOptions.java
```diff
@@ -51,28 +49,6 @@
     private Scope metricsScope;
     private boolean enableLoggingInReplay;
 
-    /**
```
vancexu, line 54: Just a reminder we may need to mention these breaking changes in next release note.

vancexu (approved):

mkolodezny (approved):

src/main/java/com/uber/cadence/worker/WorkerOptions.java
meiliang86: Yes. The next release will be 3.0

PR #552, opened 2020-10-20
Worker and Options refactoring part 2

meiliang86:
This diff mainly moved worker.factory and worker.factory.option to their own files. 
Also moved some options around.


src/main/java/com/uber/cadence/internal/worker/SingleWorkerOptions.java
```diff
@@ -29,8 +29,15 @@
 
 public final class SingleWorkerOptions {
 
-  public static final class Builder {
+  public static Builder newBuilder() {
+    return new Builder();
```
mkolodezny, line 33: then the constructor defined below must be private

src/main/java/com/uber/cadence/internal/worker/SingleWorkerOptions.java
```diff
@@ -29,8 +29,15 @@
...
+  }
 
+  public static Builder newBuilder(SingleWorkerOptions options) {
+    return new Builder(options);
```
mkolodezny, line 37: idem

mkolodezny (approved):
Just few minor comments. The overall looks good to me, great job!

src/main/java/com/uber/cadence/internal/worker/SingleWorkerOptions.java
meiliang86: Good catch. Thanks!

PR #558, opened 2020-10-24
Cleanup retryer and retry options

meiliang86:
- Renamed Retryer to RpcRetryer
- Use Default retry rpc retry option whenever possible
- Remove unnecessary retry options from worker options
- Other code cleanup


src/main/java/com/uber/cadence/internal/worker/ActivityTaskHandler.java
```diff
@@ -68,12 +66,10 @@ public Throwable getFailure() {
     public Result(
         RespondActivityTaskCompletedRequest taskCompleted,
         TaskFailedResult taskFailed,
-        RespondActivityTaskCanceledRequest taskCancelled,
-        RetryOptions requestRetryOptions) {
```
meiliang86, line 72: This is never set/used. Same for workflow task result.

src/main/java/com/uber/cadence/internal/common/WorkflowExecutionUtils.java
```diff
@@ -343,7 +333,7 @@ private static HistoryEvent getInstanceCloseEvent(
   }
 
   private static RetryOptions getRetryOptionWithTimeout(long timeout, TimeUnit unit) {
-    return new RetryOptions.Builder(retryParameters)
+    return new RetryOptions.Builder(RpcRetryer.DEFAULT_RPC_RETRY_OPTIONS)
```
vytautas-karpavicius, line 336: This line actually changes retry options, e.g. backoff coefficient is now 1.2 instead of previous 2, among others. Are we ok with this?

src/main/java/com/uber/cadence/internal/common/RpcRetryer.java
```diff
@@ -107,6 +114,10 @@ public Throwable getException() {
         });
   }
 
+  public static <T extends Throwable> void retryWithDefaultOption(RetryableProc<T> r) throws T {
```
vytautas-karpavicius, line 117: Maybe just use method overload with the same name `retry`? IMHO, this is less verbose and reads better.

src/main/java/com/uber/cadence/internal/common/WorkflowExecutionUtils.java
meiliang86: Yeah I am aware of this and I think this is ok. There is no reason that this method is special, where backoff coefficient for it is 2 instead of 1.2. Better to keep it consistent everywhere.

src/main/java/com/uber/cadence/internal/common/RpcRetryer.java
meiliang86: sure

src/main/java/com/uber/cadence/internal/common/RpcRetryer.java
```diff
@@ -57,8 +63,9 @@
         DomainAlreadyExistsError.class,
         QueryFailedError.class,
         DomainNotActiveError.class,
-        CancellationAlreadyRequestedError.class);
-    DEFAULT_SERVICE_OPERATION_RETRY_OPTIONS = roBuilder.validateBuildWithDefaults();
+        CancellationAlreadyRequestedError.class,
+        DomainNotActiveError.class);
```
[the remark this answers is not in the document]
meiliang86: Good catch. Fixed.

vytautas-karpavicius (approved):

PR #559, opened 2020-10-28
Add activity interface support

meiliang86:
Merge of this PR from Temporal: https://github.com/temporalio/sdk-java/pull/56
This is a runtime breaking change, so we need to check if we can derisk it somehow (making it backward compatible)?


shijiesheng:
close ancient PR. It's also breaking change and we will go forward with the change

PR #564, opened 2020-11-19
Add option to dispatch activity tasks locally

mkolodezny:
The activity dispatch optimization feature to allow worker to dispatch activity tasks through local tunnel after decisions are made. If the activity execution fails the retry(if any) goes the regular path(poll from server).
This is a performance optimization to skip activity scheduling efforts.

The corresponding server pr uber/cadence#3609

How did you test it?
existing integration and load tests
probably we need to update test environment as a follow up

Potential risks
No risk. It gets enabled per domain by server and off by default

### DIFF
--- a/src/main/java/com/uber/cadence/internal/metrics/MetricsType.java
+++ b/src/main/java/com/uber/cadence/internal/metrics/MetricsType.java
@@ -18,6 +18,7 @@
 package com.uber.cadence.internal.metrics;
 
 public class MetricsType {
+
   public static final String CADENCE_METRICS_PREFIX = "cadence-";
   public static final String WORKFLOW_START_COUNTER = CADENCE_METRICS_PREFIX + "workflow-start";
   public static final String WORKFLOW_COMPLETED_COUNTER =
@@ -121,7 +122,7 @@ public class MetricsType {
       CADENCE_METRICS_PREFIX + "local-activity-execution-latency";
   public static final String LOCALLY_DISPATCHED_ACTIVITY_POLL_TOTAL_COUNTER =
       CADENCE_METRICS_PREFIX + "locally-dispatched-activity-poll-total";
-  public static final String LOCALLY_DISPATCHED_ACTIVITY_POLL_NO_TASK__COUNTER =
+  public static final String LOCALLY_DISPATCHED_ACTIVITY_POLL_NO_TASK_COUNTER =
       CADENCE_METRICS_PREFIX + "locally-dispatched-activity-poll-no-task";
   public static final String LOCALLY_DISPATCHED_ACTIVITY_POLL_SUCCEED_COUNTER =
       CADENCE_METRICS_PREFIX + "locally-dispatched-activity-poll-succeed";
--- a/src/main/java/com/uber/cadence/internal/metrics/MetricsType.java
+++ b/src/main/java/com/uber/cadence/internal/metrics/MetricsType.java
@@ -119,6 +119,16 @@ public class MetricsType {
       CADENCE_METRICS_PREFIX + "local-activity-panic";
   public static final String LOCAL_ACTIVITY_EXECUTION_LATENCY =
       CADENCE_METRICS_PREFIX + "local-activity-execution-latency";
+  public static final String LOCALLY_DISPATCHED_ACTIVITY_POLL_TOTAL_COUNTER =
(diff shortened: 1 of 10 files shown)


src/main/java/com/uber/cadence/internal/metrics/MetricsType.java
```diff
@@ -119,6 +119,16 @@
       CADENCE_METRICS_PREFIX + "local-activity-panic";
   public static final String LOCAL_ACTIVITY_EXECUTION_LATENCY =
       CADENCE_METRICS_PREFIX + "local-activity-execution-latency";
+  public static final String LOCALLY_DISPATCHED_ACTIVITY_POLL_TOTAL_COUNTER =
+      CADENCE_METRICS_PREFIX + "locally-dispatched-activity-poll-total";
+  public static final String LOCALLY_DISPATCHED_ACTIVITY_POLL_NO_TASK__COUNTER =
```
vytautas-karpavicius, line 124: nit: 2 underscores `__` before `COUNTER`

src/main/java/com/uber/cadence/internal/metrics/MetricsType.java
```diff
@@ -119,6 +119,16 @@
...
+      CADENCE_METRICS_PREFIX + "locally-dispatched-activity-poll-no-task";
+  public static final String LOCALLY_DISPATCHED_ACTIVITY_POLL_SUCCEED_COUNTER =
+      CADENCE_METRICS_PREFIX + "locally-dispatched-activity-poll-succeed";
+  public static final String ACTIVITY_LOCAL_DISPATCH_FAILED_COUNTER =
+      CADENCE_METRICS_PREFIX + "activity-local-dispatch-failed";
```
vytautas-karpavicius, line 129: nit: IMHO, we should keep consistent suffixes, either both -fail and -succeed or both fail**ed** and succeed**ed**

src/main/java/com/uber/cadence/internal/sync/SyncWorkflowWorker.java
```diff
@@ -115,64 +136,80 @@ public void setLocalActivitiesImplementation(Object... activitiesImplementation)
...
+    return workflowWorker.isTerminated()
+        && laWorker.isTerminated()
+        && ldaHeartbeatExecutor.isTerminated()
+        && ldaWorker.isTerminated();
   }
 
   @Override
   public void shutdown() {
     laWorker.shutdown();
+    ldaHeartbeatExecutor.shutdown();
```
vytautas-karpavicius, line 175: Why are we shutting down `ldaHeartbeatExecutor` but not the `heartbeatExecutor`? How do they differ?

src/main/java/com/uber/cadence/internal/sync/SyncActivityWorker.java
```diff
@@ -37,7 +37,9 @@ public SyncActivityWorker(
       IWorkflowService service, String domain, String taskList, SingleWorkerOptions options) {
     taskHandler =
         new POJOActivityTaskHandler(service, domain, options.getDataConverter(), heartbeatExecutor);
-    worker = new ActivityWorker(service, domain, taskList, options, taskHandler);
+    worker =
+        new ActivityWorker(
+            service, domain, taskList, options, taskHandler, "Activity Poller taskList=");
```
vytautas-karpavicius, line 42: IMO, `taskList=` part should belong inside ActivityWorker implementation.

vytautas-karpavicius (approved):
Overall looks good! Left several comments but nothing major.

src/main/java/com/uber/cadence/internal/sync/SyncActivityWorker.java
mkolodezny: good catch, agree!

src/main/java/com/uber/cadence/internal/sync/SyncWorkflowWorker.java
mkolodezny: yeah, I had decided to test it in the separate diff to be safe.

src/main/java/com/uber/cadence/internal/metrics/MetricsType.java
mkolodezny: agree, but that should be refactored at the same time with go client tho. Also there are alerts and customers who have already sticked to those constants.

PR #567, opened 2020-12-13
do not dispatch activities locally if taskListActivitiesPerSecond is set

mkolodezny:


src/main/java/com/uber/cadence/internal/sync/SyncWorkflowWorker.java
```diff
@@ -147,69 +157,84 @@ public void start() {
     // to start LocalActivity Worker.
     if (workflowWorker.isStarted()) {
       laWorker.start();
-      ldaWorker.start();
+      if (ldaWorker != null) {
```
vytautas-karpavicius, line 160: Have you considered using [NoopSuspendableWorker](https://github.com/uber/cadence-java-client/blob/master/src/main/java/com/uber/cadence/internal/worker/NoopSuspendableWorker.java) (or a variant of it) for non initialized worker?
This would eliminate all those null checks.

vytautas-karpavicius (approved):

longquanzheng:
@mkolodezny @vytautas-karpavicius Hello, I am curious why this change? is any issue with taskListActivitiesPerSecond and local dispatch?

PR #568, opened 2020-12-15
Adjust various default timeout values

longquanzheng:
1. Increase default RPC timeout from 1 second to 5 seconds to avoid unnecessary retry in prod env
2. Decrease default long poll timeout from 125 seconds to 20 seconds to reduce down time when restarting server
3. Increase buildkite timeout from 15 mins to 30 minutes to fix buildkite failure


src/main/java/com/uber/cadence/serviceclient/ClientOptions.java
```diff
@@ -29,9 +29,9 @@
   private static final String LOCALHOST = "127.0.0.1";
 
   /** Default RPC timeout used for all non long poll calls. */
-  private static final long DEFAULT_RPC_TIMEOUT_MILLIS = 1000;
+  private static final long DEFAULT_RPC_TIMEOUT_MILLIS = 5 * 1000;
```
longquanzheng, line 32: change to 3 secs

meiliang86 (approved):

meiliang86 (approved):

Issue #565, opened 2020-11-27
Possible racing condition bug in history iterator


longquanzheng:
Cannot be reproduced stably.
Version: 2.7.1
![stacktrace](https://user-images.githubusercontent.com/4523955/100399683-930e6d80-3008-11eb-82bf-d865d358f2fc.png)
![code-trace](https://user-images.githubusercontent.com/4523955/100399686-94d83100-3008-11eb-86f8-1fa3f3265b6a.png)

https://uber-cadence.slack.com/archives/CM29NND28/p1606398004176000

longquanzheng:
https://github.com/uber/cadence-java-client/pull/574

PR #576, opened 2020-12-22
Prevent iterator panic and add error log

longquanzheng:
For #565
Per discussion in slack channel, we want to add a check and error log for this case.

### DIFF
--- a/src/main/java/com/uber/cadence/internal/replay/ReplayDecider.java
+++ b/src/main/java/com/uber/cadence/internal/replay/ReplayDecider.java
@@ -684,7 +684,7 @@ class ReplayDecider implements Decider {
           }
           if (!current.hasNext()) {
             log.error(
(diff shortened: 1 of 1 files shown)


meiliang86 (approved):

src/main/java/com/uber/cadence/internal/replay/ReplayDecider.java
```diff
@@ -677,6 +682,11 @@ public HistoryEvent next() {
             metricsScope.counter(MetricsType.WORKFLOW_GET_HISTORY_FAILED_COUNTER).inc(1);
             throw new Error(e);
           }
+          if (!current.hasNext()) {
+            log.error("GetWorkflowExecutionHistory return empty history, maybe a bug in server, token:"
```
meiliang86, line 686: Add workflowId, runID and domainID in the log.

PR #582, opened 2021-02-01
Clean up workflow query handling

meiliang86:
There are two major changes in this PR:
1. Allow user to set QueryRejectionCondition via WorkflowClientOptions, so that user can use QueryRejectionCondition on strongly typed QueryMethods.
2. Remove QueryResult in favor of WorkflowQueryRejectedException.

Ref to similar changes in Temporal: https://github.com/temporalio/sdk-java/pull/69


src/main/java/com/uber/cadence/client/WorkflowClientOptions.java
```diff
@@ -175,4 +200,52 @@ public String getIdentity() {
...
+        + '\''
+        + ", contextPropagators="
+        + contextPropagators
+        + ", queryRejectCondition="
+        + queryRejectCondition
+        + '}';
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
vytautas-karpavicius, line 229: It looks that `equals` and `hashCode` are not used anywhere. Do we really need those methods?

vytautas-karpavicius (approved):

andrewjdawson2016 (approved):

PR #587, opened 2021-02-27
Add shadowing options for the worker shadowing mode

yux0:
Add shadowing options for the worker shadowing mode.

### DIFF
--- a/src/main/java/com/uber/cadence/worker/ShadowingExitCondition.java
+++ b/src/main/java/com/uber/cadence/worker/ShadowingExitCondition.java
@@ -1,5 +1,5 @@
 /*
- *  Modifications Copyright (c) 2017-2020 Uber Technologies Inc.
+ *  Modifications Copyright (c) 2017-2021 Uber Technologies Inc.
  *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  *  Licensed under the Apache License, Version 2.0 (the "License"). You may not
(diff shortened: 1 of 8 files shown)


src/main/java/com/uber/cadence/worker/ShadowingExitCondition.java
```diff
@@ -0,0 +1,99 @@
+/*
+ *  Modifications Copyright (c) 2017-2020 Uber Technologies Inc.
```
yycptt, line 2: nit: 2021

src/main/java/com/uber/cadence/worker/ShadowingOptions.java
```diff
@@ -0,0 +1,188 @@
...
+            Objects.requireNonNull(workflowStatuses);
+            this.workflowStatuses = Sets.newHashSet(workflowStatuses);
+            if (workflowStatuses.size() == 0) {
+                this.workflowStatuses.add(WorkflowStatus.OPEN);
+            }
+            return this;
+        }
+
+        public Builder setWorkflowSamplingRate(double samplingRate) {
+            if (samplingRate < 0.0 || samplingRate > 1.0) {
```
yycptt, line 94: nit: shall we also treat 0 as invalid?

src/main/java/com/uber/cadence/worker/TimeFilter.java
```diff
@@ -0,0 +1,87 @@
...
+        return DEFAULT_INSTANCE;
+    }
+
+    private static final TimeFilter DEFAULT_INSTANCE;
+
+    static {
+        DEFAULT_INSTANCE = TimeFilter.newBuilder().build();
+    }
+
+    public static final class Builder {
```
yycptt, line 41: nit: I am not sure if having a Builder for every option field is an overkill or not. But I am not familiar with java client, so feel free to ignore this comment.

yycptt (approved):
LGTM. 

We may need to think about how to reuse part of the code when defining the options for shadowing worker. I guess it will be very similar to this option for local shadowing?

src/main/java/com/uber/cadence/worker/ShadowingOptions.java
yux0: Yes. Updated

src/main/java/com/uber/cadence/worker/TimeFilter.java
yux0: Had the same feeling. But I incline to keep the consistency here.

PR #588, opened 2021-03-02
Add query builder for shadowing options

yux0:
Add query builder for shadowing options

### DIFF
--- a/src/main/java/com/uber/cadence/internal/shadowing/QueryBuilder.java
+++ b/src/main/java/com/uber/cadence/internal/shadowing/QueryBuilder.java
@@ -18,6 +18,7 @@ package com.uber.cadence.internal.shadowing;
 import com.google.common.collect.Lists;
 import com.uber.cadence.worker.TimeFilter;
 import com.uber.cadence.worker.WorkflowStatus;
+import java.time.ZonedDateTime;
 import java.util.Collection;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/uber/cadence/internal/shadowing/QueryBuilder.java
```diff
@@ -0,0 +1,121 @@
...
+    this.stringBuffer = new StringBuffer();
+  }
+
+  private void appendPartialQuery(String query) {
+    if (query == null || query.length() == 0) {
+      return;
+    }
+
+    if (this.stringBuffer.length() != 0) {
+      this.stringBuffer.append(" and ");
```
yycptt, line 114: nit: use `AND_QUERY`?

src/main/java/com/uber/cadence/internal/shadowing/QueryBuilder.java
```diff
@@ -0,0 +1,121 @@
...
+
+  public QueryBuilder setWorkflowStartTime(TimeFilter timeFilter) {
+    if (timeFilter == null) {
+      return this;
+    }
+
+    Collection<String> timerFilters = Lists.newArrayListWithCapacity(2);
+    if (timeFilter.getMinTimestamp() != null) {
+      timerFilters.add(
+          START_TIME_PLACEHOLDER + " >= " + timeFilter.getMinTimestamp().toEpochSecond());
```
yycptt, line 78: should be nano second?

src/main/java/com/uber/cadence/internal/shadowing/QueryBuilder.java
yux0: updated

src/main/java/com/uber/cadence/internal/shadowing/QueryBuilder.java
yux0: Yes. Updated with a get nano function

yycptt (approved):

PR #590, opened 2021-03-02
Add replay activity to replay workflow history

yux0:
Add replay activity to replay workflow history


src/main/java/com/uber/cadence/internal/shadowing/ReplayWorkflowActivityImpl.java
```diff
@@ -0,0 +1,133 @@
...
+                successCount, failedCount, skippedCount, failedExecution);
+          }
+        } else {
+          this.metricsScope.counter(MetricsType.SHADOWING_SKIPPED_COUNT).inc(1);
+          skippedCount++;
+        }
+        continue;
+      }
+      this.metricsScope.counter(MetricsType.SHADOWING_SUCCESS_COUNT).inc(1);
+      successCount++;
```
yycptt, line 110: shall we use heartbeat to record the current progress?

src/main/java/com/uber/cadence/internal/shadowing/ReplayWorkflowActivityResponse.java
```diff
@@ -0,0 +1,53 @@
...
+package com.uber.cadence.internal.shadowing;
+
+import com.uber.cadence.WorkflowExecution;
+import java.util.List;
+
+public class ReplayWorkflowActivityResponse {
+  private final int successCount;
+  private final int failureCount;
+  private final int skippedCount;
+  private final List<WorkflowExecution> failedWorkflowExecutions;
```
yycptt, line 25: let's have a sync offline to make the naming consistent with go-client?

src/main/java/com/uber/cadence/internal/shadowing/ReplayWorkflowActivityImpl.java
```diff
@@ -0,0 +1,133 @@
...
+  private final Class<?>[] workflowClasses;
+
+  public ReplayWorkflowActivityImpl(
+      IWorkflowService serviceClient,
+      boolean failFast,
+      Scope metricsScope,
+      Class<?> workflowClass,
+      Class<?>... workflowClasses) {
+    this.serviceClient = serviceClient;
+    this.failFast = failFast;
```
yycptt, line 50: Is it for local test? If so, shall we also include exit condition for shadowing?

yux0: Removed this flag

src/main/java/com/uber/cadence/internal/shadowing/ReplayWorkflowActivityResponse.java
yux0: removed the list of failed workflow executions

src/main/java/com/uber/cadence/internal/shadowing/ReplayWorkflowActivityImpl.java
yux0: Yes. Updated

src/main/java/com/uber/cadence/internal/shadowing/ReplayWorkflowActivityImpl.java
```diff
@@ -0,0 +1,225 @@
...
+        .setSucceeded(successCount)
+        .setFailed(failedCount)
+        .setSkipped(skippedCount);
+  }
+
+  public ReplayWorkflowActivityResult replayHelper(String domain, WorkflowExecution execution) {
+    WorkflowExecutionHistory workflowHistory;
+    try {
+      workflowHistory = getFullHistory(domain, execution);
+    } catch (Exception e) {
```
yycptt, line 110: for workflow type not defined error, shall we failed the workflow instead of skipping it? I think it usually means there's a something wrong with workflow registration and require a worker deployment to fix it. 

In go client, I returned a non-retryable error defined in thrift for the activity so that the workflow will fail. What's your opinion?

yux0: I am thinking of it as a replay failure and emit some replay metrics. The use case in my mind is, for example, there are 2 wf types. one is registered and one doesn't. We should replay for those 2 wf types and emit failure count on the unregister ones. What do you think?

src/main/java/com/uber/cadence/internal/shadowing/ReplayWorkflowActivityImpl.java
```diff
@@ -0,0 +1,225 @@
...
+        throw e;
+      } else {
+        log.info("replay workflow skipped", e);
+        return false;
+      }
+    } finally {
+      sw.stop();
+    }
+
+    log.info("replay workflow succeed");
```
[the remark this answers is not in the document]
yux0: updated

yycptt (approved):

PR #595, opened 2021-03-17
Add Shadow worker

yux0:
Add Shadow worker.


yycptt (approved):
LGTM

src/main/java/com/uber/cadence/worker/ShadowingWorker.java
```diff
@@ -0,0 +1,180 @@
...
+        new SyncActivityWorker(
+            client.getService(), client.getOptions().getDomain(), this.taskList, activityOptions);
+    activityWorker.setActivitiesImplementation(scanActivity, replayActivity);
+  }
+
+  void start() throws Exception {
+    if (!started.compareAndSet(false, true)) {
+      return;
+    }
+    startShadowingWorkflow();
```
yycptt, line 93: shall we validate shadow options here? e.g. make sure exit condition is set when using continuous mode.

src/main/java/com/uber/cadence/worker/ShadowingWorker.java
```diff
@@ -0,0 +1,180 @@
...
+            .setDomain(shadowingOptions.getDomain())
+            .setConcurrency(shadowingOptions.getConcurrency())
+            .setExitCondition(shadowingOptions.getExitCondition())
+            .setShadowMode(shadowingOptions.getShadowMode())
+            .setSamplingRate(shadowingOptions.getSamplingRate())
+            .setTaskList(taskList)
+            .setWorkflowQuery(query);
+    StartWorkflowExecutionRequest request =
+        new StartWorkflowExecutionRequest()
+            .setDomain(shadowingOptions.getDomain())
```
yycptt, line 168: this should be our system domain

yycptt (changes requested):

src/main/java/com/uber/cadence/worker/ShadowingWorker.java
yux0: Good catch. updated

src/main/java/com/uber/cadence/worker/ShadowingWorker.java
yux0: Added into options build()

yycptt (approved):

PR #596, opened 2021-03-18
Add workflow shadower for local shadow testing

yux0:
Introduce a local testing wf shadower to allow shadowing prod traffic.


src/main/java/com/uber/cadence/testing/WorkflowShadower.java
```diff
@@ -0,0 +1,118 @@
...
+
+        // Check exit condition
+        if (replayCount >= maxReplayCount) {
+          return;
+        }
+        if (ZonedDateTime.now().isAfter(now.plus(maxReplayDuration))) {
+          return;
+        }
+      }
+    } while (nextPageToken != null);
```
yycptt, line 116: Do we plan to support different shadow mode?

yux0: yes. added it.

src/main/java/com/uber/cadence/testing/WorkflowShadower.java
```diff
@@ -0,0 +1,116 @@
...
+        // Check exit condition
+        if (replayCount >= maxReplayCount) {
+          return;
+        }
+        if (!maxReplayDuration.isZero()
+            && ZonedDateTime.now().isAfter(now.plusSeconds(maxReplayDuration.getSeconds()))) {
+          return;
+        }
+      }
+    } while (nextPageToken != null && options.getShadowMode() == Mode.Normal);
```
yycptt, line 114: add wait period between two scan iterations?

src/main/java/com/uber/cadence/worker/ShadowingOptions.java
```diff
@@ -162,7 +163,14 @@ public ShadowingOptions build() {
         throw new IllegalArgumentException(
             "workflow types, status and start time filter can't be specified when workflow query is specified");
       }
-
+      if (workflowQuery.isEmpty()) {
```
yycptt, line 166: I probably missed this in the previous diff. Would you mind provide some context behind checking for OPEN status on L161-162?

src/main/java/com/uber/cadence/testing/WorkflowShadower.java
```diff
@@ -0,0 +1,116 @@
...
+import java.time.ZonedDateTime;
+
+public final class WorkflowShadower {
+
+  private final ShadowingOptions options;
+  private final String query;
+  private final ScanWorkflowActivity scanWorkflow;
+  private final ReplayWorkflowActivity replayWorkflow;
+
+  public WorkflowShadower(IWorkflowService service, ShadowingOptions options, String taskList) {
```
yycptt, line 43: I think we don't support shadow options with concurrency > 1 for local shadower. Is there a check for it and will an exception be raised?

src/main/java/com/uber/cadence/testing/WorkflowShadower.java
```diff
@@ -0,0 +1,116 @@
...
+              new ScanWorkflowActivityParams()
+                  .setDomain(options.getDomain())
+                  .setWorkflowQuery(query)
+                  .setNextPageToken(nextPageToken)
+                  .setSamplingRate(options.getSamplingRate()));
+      nextPageToken = ScanResult.getNextPageToken();
+
+      for (WorkflowExecution execution : ScanResult.getExecutions()) {
+        ReplayWorkflowActivityResult replayResult =
+            replayWorkflow.replay(
```
yycptt, line 94: Is there any logging inside the replay activity for why the replay failed or skipped?

src/main/java/com/uber/cadence/worker/ShadowingOptions.java
yux0: The default workflow status is OPEN. Here it checks if both query  and filter are setup

src/main/java/com/uber/cadence/testing/WorkflowShadower.java
yux0: Added

src/main/java/com/uber/cadence/testing/WorkflowShadower.java
yux0: yes

src/main/java/com/uber/cadence/testing/WorkflowShadower.java
yux0: add a check

src/main/java/com/uber/cadence/testing/WorkflowShadower.java
yycptt: hmmm, looks like this sleep interval will also be applied with in an iteration, which is not what we want?

src/main/java/com/uber/cadence/worker/ShadowingOptions.java
yycptt: oh, I see. I think I missed this part in go client's implementation. Right now in go, if no status is specified it will query all workflows.

yux0: I see. I set the default to be OPEN workflows

src/main/java/com/uber/cadence/testing/WorkflowShadower.java
yux0: Good point. Updated

yycptt (approved):

PR #601, opened 2021-04-08
Add task list type for normal and sticky task list

yux0:
What?
Add task list type for normal and sticky task list

Why?
Sticky tasklists are random generated and causes noise in metics.


meiliang86:
Can you check if we do the same in Go? We should make the behavior consistent.

yux0:
> Can you check if we do the same in Go? We should make the behavior consistent.

Change back to use a placeholder for sticky tasklist. Will check the same in go client

Groxx (approved):
Sorry about the delay!

yeah, I think this is probably correct.  Server seems to use `__sticky__`, so clients should too: https://github.com/uber/cadence/blob/7bd91051360214181464b87325615631a1712bb7/service/matching/context.go#L56-L75

PR #603, opened 2021-04-13
Add register workflow methods in shadower

yux0:
1. What change?
Add register workflow methods in shadower

2. Why?
User can register the workflow instead of pass the activity instance in.


src/main/java/com/uber/cadence/testing/WorkflowShadower.java
```diff
@@ -114,6 +116,28 @@ public void run() throws Throwable {
     } while (nextPageToken != null && options.getShadowMode() == Mode.Normal);
   }
 
+  public void registerWorkflowImplementationTypes(Class<?>... workflowImplementationClasses) {
```
yycptt, line 119: I wonder how does the test work if we can't register workflow to the shadower?

yycptt (approved):

src/main/java/com/uber/cadence/testing/WorkflowShadower.java
yux0: In the test, I register workflows in the replay activity, then inject the replay activity to the shadower. Users doesn't need to know the activity.

PR #604, opened 2021-04-15
Expose test environment option for shadowing

yux0:
What: 
Expose test environment option for shadowing

Why: 
User can pass the data converter, interceptor, contextPropagator to replay test worker.


yycptt (approved):
LGTM! Is it possible to also add a unit test for replay that uses one of the worker options like data converter or context propagator, etc. ?

yux0:
> LGTM! Is it possible to also add a unit test for replay that uses one of the worker options like data converter or context propagator, etc. ?

Yes. Modified the replay tests.

PR #606, opened 2021-04-16
Support workflow completed error in java client

demirkayaender:
**What changed?**
Similar changes to here: https://github.com/uber-go/cadence-client/pull/1081 for the java client. 

**Why?**

**How did you test it?**

**Potential risks**


yux0 (commented):
LGTM. Could we add some tests to make sure this new errors are not retry?
As this is a breaking change, may be update the version in internal/Version.java as well

yux0 (approved):

demirkayaender:
**What changed?**
Similar changes to here: https://github.com/uber-go/cadence-client/pull/1081 for the java client. 


**Why?**



**How did you test it?**


**Potential risks**

PR #607, opened 2021-04-19
Add activity worker thread count metrics

Shaddoll:
What?

Add activity worker thread count metrics to track the number of concurrent activities in a worker.

### DIFF
--- a/src/main/java/com/uber/cadence/internal/sync/POJOActivityTaskHandler.java
+++ b/src/main/java/com/uber/cadence/internal/sync/POJOActivityTaskHandler.java
@@ -19,6 +19,7 @@ package com.uber.cadence.internal.sync;
 
 import com.google.common.base.Joiner;
 import com.google.common.reflect.TypeToken;
+import com.google.common.util.concurrent.RateLimiter;
 import com.uber.cadence.PollForActivityTaskResponse;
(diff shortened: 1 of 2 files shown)


src/main/java/com/uber/cadence/internal/sync/POJOActivityTaskHandler.java
```diff
@@ -186,6 +188,9 @@ public Result handle(
           metricsScope,
           isLocalActivity);
     }
+    if (metricsRateLimiter.tryAcquire(1)) {
+      metricsScope.gauge(MetricsType.ACTIVITY_ACTIVE_THREAD_COUNT).update(Thread.activeCount());
```
meiliang86, line 192: Add a tag to indicate local v.s. normal activity.

Shaddoll: I use a different metrics type for local activity to match the existing metrics naming pattern.
ACTIVITY_ACTIVE_THREAD_COUNT is the thread count of normal activity.
LOCAL_ACTIVITY_ACTIVE_THREAD_COUNT is the thread count of local activity.

meiliang86 (approved):

PR #608, opened 2021-04-20
add all missing version marker before upsert searchattributes

Shaddoll:
why?

We allow user to remove `getversion` call in their workflows. Current implementation throws a non-deterministic error if the `getversion` call before `upsertsearchattributes` is removed

test?
new unit test


meiliang86:
Add a unit test (maybe a replay test)?

meiliang86:
Check WorkflowTests.java. You should find existing tests that you can borrow.

meiliang86 (approved):

PR #617, opened 2021-06-15
Health check

AngerM-DD:
This adds a health check to the cadence-java sdk. It is useful is k8s environments where you pass in a DNS name as opposed to an IP address. This change allows you to detect if the client can no longer connect to the cadence servers and attempt to re-resolve the dns name if one was used.


src/main/java/com/uber/cadence/serviceclient/WorkflowServiceTChannel.java
```diff
@@ -207,6 +215,45 @@ private static String getEndpoint(String service, String method) {
     return buildThriftRequest(apiName, body, null);
   }
 
+  @Override
+  public CompletableFuture<Boolean> isHealthy() {
```
longquanzheng, line 219: The implementation is doing more than the name. Do we have a better way to describe it? 

Alternatively, I think this should be two separate methods: `isHeathy` and `resetEndpoint`

src/main/java/com/uber/cadence/serviceclient/WorkflowServiceTChannel.java
```diff
@@ -207,6 +215,45 @@ private static String getEndpoint(String service, String method) {
...
+    final ThriftRequest<Meta.health_args> req =
+        new ThriftRequest.Builder<Meta.health_args>("cadence-frontend", "Meta::health")
```
longquanzheng, line 221: I think hardcoding `"cadence-frontend"` will not work for everyone. `cadence-frontend` just happen to be the default service name for frontend service. For example, in Uber because there are multiple cadence clusters, they have to name it `cadence-frontend-clusterA` `cadence-frontend-clusterB`.  Is this just required for `Meta::health`. How does other TChannel method works?

src/main/java/com/uber/cadence/worker/WorkerFactory.java
```diff
@@ -287,6 +289,16 @@ public synchronized void shutdownNow() {
     }
   }
 
+  public CompletableFuture<Boolean> isHealthy() {
```
longquanzheng, line 292: Does it mean that the application will need to call this periodically?

src/main/java/com/uber/cadence/serviceclient/WorkflowServiceTChannel.java
```diff
@@ -237,6 +284,11 @@ private static String getEndpoint(String service, String method) {
     } catch (ExecutionException e) {
       throw new TException(e);
     } catch (TChannelError e) {
+      try {
+        resetSubchannelPeers();
```
longquanzheng, line 288: I think it's better not reset here. The latency of API will become unpredictable with this change. Maybe worker/workerFactory could have a scheduledExecutionService to do heathcheck and reset endpont every 10s.

src/main/java/com/uber/cadence/worker/WorkerFactory.java
AngerM-DD: Yes. Our clients call this as part of their k8s health check endpoints

longquanzheng: I see. It's better to have scheduledExecutionService to do the work automatically. This is a nice improvement, but many users don't need to understand this.

src/main/java/com/uber/cadence/serviceclient/WorkflowServiceTChannel.java
AngerM-DD: Ok. We added it everywhere to try to reduce the amount of 'downtime' a client would be unable to talk to the cluster

longquanzheng: I agreed it would work in most times. But still not a good practice to do for this API. 
By contract, each service API in this component should do only one thing and sending request to the service. 

Having a background process to reset should be good enough. If 10s is too long for downtime, make it configurable and even default to 1 second. Heathcheck for every 1s is fine. 

We should add some metrics to this background process for visibility. In case we see a worker constatnly resetting endpoint, there is something wrong elsewhere. 

Not only additional latency, this `TChannelError` may not be safe for resetting endpoint for all service APIs, which may end up resetting for every request.

src/main/java/com/uber/cadence/serviceclient/WorkflowServiceTChannel.java
AngerM-DD: `resetSubchannelPeers()` is already public. The reason why we combined them is to make it a one stop shop

longquanzheng: Better to call it `resetEndpoint`. The interface is not only for TChannel. gRPC is on the way. 

Combining it doesn't sound great with the name, and is unnecessary effort. I would keep it separate and let the caller to decide how to do with healthcheck return false.

AngerM-DD: grpc doesn't (or shouldn't) have this issue, since the underlying library will re resolve the dns if there is a connection failure for you. tchannel doesn't understand DNS which is why we added this functionality. Assuming Cadence's GRPC integration is similar to Temporals, the resets won't be necessary for tchannel.

longquanzheng: I see. so looks like that's why you only put in WorkflowServiceTChannel not in the interface. 

I am fine with just expose IsHealthy()

src/main/java/com/uber/cadence/serviceclient/IWorkflowService.java
```diff
@@ -84,4 +86,6 @@ void SignalWorkflowExecutionWithTimeout(
       AsyncMethodCallback resultHandler,
       Long timeoutInMillis)
       throws TException;
+
+  CompletableFuture<Boolean> isHealthy();
```
longquanzheng, line 90: I can't think of a new name and kinda like `isHeathy`. So can you add a comment about the fact that implementation may try to resolve the unhealthy issue like reset DNS endpoint? Otherwise LGTM

longquanzheng (approved):
LGTM(except that the name of `isHealthy` may be misleading for caller. 

@mkolodezny @meiliang86 do you also want to take a look?
This PR should be safe to land -- it introduces a new optional API in worker/workerFactory and doesn't change any existing behavior.

@AngerM-DD thank you for the contribution.

src/main/java/com/uber/cadence/worker/WorkerFactory.java
```diff
@@ -287,6 +289,16 @@ public synchronized void shutdownNow() {
     }
   }
```
longquanzheng, line 291: @AngerM-DD the comment I mentioned should also be put here  and the IWorkflowService for clarification. As these are the two places that users could look at.

src/main/java/com/uber/cadence/serviceclient/IWorkflowService.java
AngerM-DD: Do you have a better idea for a name?

longquanzheng (approved):

PR #618, opened 2021-06-15
Refactor context prop support to allow things like opentelemtry to work properly with the cadence sdk

AngerM-DD:
Refactor the context propagators to support OpenTelemetry and add an example OTEL propagator.

### DIFF
--- a/src/main/java/com/uber/cadence/context/ContextPropagator.java
+++ b/src/main/java/com/uber/cadence/context/ContextPropagator.java
@@ -23,6 +23,9 @@ import java.util.Map;
  * Context Propagators are used to propagate information from workflow to activity, workflow to
  * child workflow, and workflow to child thread (using {@link com.uber.cadence.workflow.Async}).
  *
+ * <p>It is important to note that all threads share one ContextPropagator instance, so your
+ * implementation <b>must</b> be thread-safe and store any state in ThreadLocal variables.
+ *
  * <p>A sample <code>ContextPropagator</code> that copies all {@link org.slf4j.MDC} entries starting
  * with a given prefix along the code path looks like this:
  *
@@ -136,4 +139,31 @@ public interface ContextPropagator {
 
   /** Sets the current context */
   void setCurrentContext(Object context);
+
+  /**
+   * This is a lifecycle method, called after the context has been propagated to the
+   * workflow/activity thread but the workflow/activity has not yet started.
+   */
+  default void setUp() {
+    // No-op
+  }
+
+  /**
+   * This is a lifecycle method, called after the workflow/activity has completed. If the method
+   * finished without exception, {@code successful} will be true. Otherwise, it will be false and
+   * {@link #onError(Throwable)} will have already been called.
+   */
+  default void finish() {
+    // No-op
+  }
+
+  /**
+   * This is a lifecycle method, called when the workflow/activity finishes by throwing an unhandled
+   * exception. {@link #finish()} is called after this method.
+   *
+   * @param t The unhandled exception that caused the workflow/activity to terminate
+   */
+  default void onError(Throwable t) {
+    // No-op
+  }
 }
--- a/src/main/java/com/uber/cadence/context/OpenTelemetryContextPropagator.java
+++ b/src/main/java/com/uber/cadence/context/OpenTelemetryContextPropagator.java
@@ -1,0 +1,144 @@
+/*
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"). You may not
+ *  use this file except in compliance with the License. A copy of the License is
+ *  located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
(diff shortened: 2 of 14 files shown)


src/main/java/com/uber/cadence/context/OpenTelemetryContextPropagator.java
```diff
@@ -0,0 +1,144 @@
...
+        Baggage.fromContext(w3cBaggagePropagator.extract(Context.current(), null, getter))
+            .toBuilder()
+            .build()
+            .storeInContext(w3cTraceContextPropagator.extract(Context.current(), null, getter));
+
+    currentContextOtelScope.set(context.makeCurrent());
+
+    Span span =
+        GlobalOpenTelemetry.getTracer("cadence-client")
+            .spanBuilder("cadence.workflow")
```
longquanzheng, line 119: Not familar with opentelemtry -- How are these two properties used? I wonder if we should avoid harhcoding here because users may need to customize them.

src/main/java/com/uber/cadence/context/ContextPropagator.java
```diff
@@ -136,4 +139,31 @@
...
+   * This is a lifecycle method, called after the context has been propagated to the
+   * workflow/activity thread but the workflow/activity has not yet started.
+   */
+  default void setUp() {
+    // No-op
+  }
+
+  /**
+   * This is a lifecycle method, called after the workflow/activity has completed. If the method
+   * finished without exception, {@code successful} will be true. Otherwise, it will be false and
```
longquanzheng, line 153: Where/what is `{@code successful}`?

src/main/java/com/uber/cadence/context/ContextPropagator.java
```diff
@@ -136,4 +139,31 @@
...
+   */
+  default void finish() {
+    // No-op
+  }
+
+  /**
+   * This is a lifecycle method, called when the workflow/activity finishes by throwing an unhandled
+   * exception. {@link #finish()} is called after this method.
+   *
+   * @param t The unhandled exception that caused the workflow/activity to terminate
```
longquanzheng, line 164: Nit: `terminate` is a special in Cadence only for specific cases of closed. You can use either `close` or `stop` to describe different cases finish with exceptions

demirkayaender:
I triggered a build for this PR but as with my other PR (https://github.com/uber/cadence-java-client/pull/619), this one is failing build too. Something is broken with Buildkite and we are trying to understand what the issue is but `compileThrift` step is failing for some reason. We don't know the exact reason yet but hoping to resolve it soon. Let us know if you can find out something to fix the issue.

demirkayaender:
Some update here: The issue is related to Java11 and Alpine. We are looking into how we could fix it. All the PRs will be potentially blocked until the fix unfortunately.

demirkayaender:
Fix is here: https://github.com/uber/cadence-java-client/pull/625

src/main/java/com/uber/cadence/context/OpenTelemetryContextPropagator.java
```diff
@@ -0,0 +1,144 @@
...
+    Span span = currentOtelSpan.get();
+    if (span != null) {
+      span.end();
+    }
+    Scope contextScope = currentContextOtelScope.get();
+    if (contextScope != null) {
+      contextScope.close();
+    }
+  }
+}
```
longquanzheng, line 144: Why the `onError` is not implemented? is it not needed for OpenTelemetry?

src/main/java/com/uber/cadence/internal/sync/WorkflowStubImpl.java
```diff
@@ -202,7 +203,18 @@ private StartWorkflowExecutionParameters getStartWorkflowExecutionParameters(
...
+          propagator.serializeContext(propagator.getCurrentContext());
+      // Namespace each entry in case of overlaps, so foo -> bar becomes MyPropagator:foo -> bar
+      Map<String, byte[]> namespacedSerializedContext =
+          serializedContext
+              .entrySet()
+              .stream()
+              .collect(
+                  Collectors.toMap(
+                      k -> propagator.getName() + ":" + k.getKey(), Map.Entry::getValue));
+      result.putAll(namespacedSerializedContext);
```
longquanzheng, line 217: maybe move to a util to reuse the code. It's repeated 4 times for ChildWF, WF, activity and localActivity.

src/main/java/com/uber/cadence/internal/replay/WorkflowContext.java
```diff
@@ -166,7 +167,18 @@ SearchAttributes getSearchAttributes() {
...
+      Map<String, byte[]> filteredData =
+          headerData
+              .entrySet()
+              .stream()
+              .filter(e -> e.getKey().startsWith(propagator.getName()))
+              .collect(
+                  Collectors.toMap(
+                      e -> e.getKey().substring(propagator.getName().length() + 1),
+                      Map.Entry::getValue));
+      contextData.put(propagator.getName(), propagator.deserializeContext(filteredData));
```
longquanzheng, line 181: This is nice. Looks like the existing one is having issues with more than one ContextPropagators, right? @meiliang86 @mkolodezny 

However, I believe there is a backward-compatibility issue by doing it. 
The existing workflow using ContextPropagator already made headers without this format...
To be backward-compatible, maybe we should add an option to use this and default to false.

longquanzheng: Another idea is that we can safely assume the existing propagators don't start with some special string as key(e.g. `__cadence_v1_ctx_key` ) and then we can check if the headers have any keys with it. 

If yes then we can use this new way. If not, then fallback to use the old way.

shijiesheng:
We are not committed to make this big change yet and I've changed it to an issue https://github.com/cadence-workflow/cadence-java-client/issues/1054.

If you are still interested in contributing, please feel free to reopen it.

PR #619, opened 2021-06-18
Support feature flags in java client

demirkayaender:
**What changed?**
Similar to https://github.com/uber-go/cadence-client/pull/1103 change in go client, this PR introduces feature flags to java client so customers using java can enable/disable some features.

**Why?**
Without this change, customers won't be able to turn on some features.

**How did you test it?**
```
USE_DOCKER_SERVICE=true ./gradlew test --tests *testSignalingCompletedWorkflow* // with real server

./gradlew test --tests *testSignalingCompletedWorkflow*
```

Also tested with java samples

**Potential risks**


demirkayaender:
Failure is from the integration test using older server Docker images which doesn't support the new error. On Monday, I will ask how to update the server version.

emrahs (approved):

PR #624, opened 2021-07-13
Revert "Add Health check API to worker and service interface (#617)"

demirkayaender:
This reverts commit 63a34bc3238a6b6a84cbb9d79f564bc8ae311a37.


longquanzheng:
Is that heathcheck causing some issues? I thought it’s all optional and backward compatibility. So maybe I missed something

demirkayaender:
> Is that heathcheck causing some issues? I thought it’s all optional and backward compatibility. So maybe I missed something

Please ignore these PRs, I will abandon this. There's a build failure that only happens in Buildkite, I was sending PRs to trigger the build and see if I could pinpoint what caused it.

PR #625, opened 2021-07-13
Fix Java Client Build

demirkayaender:
Java client started using java11:alpine in April. Since we were using the branch instead of a specific commit, we have been getting updates for jdk. One of those updates 15 days ago ended up breaking the build and all the subsequent builds failed: https://buildkite.com/uberopensource/cadence-java-client/builds?branch=master&page=2 

This change freezes the jdk version to a specific hash so we won't have similar breakages in the future.

I also switched to using paths with no variables like $projectDir; this was because usage of it initially mislead us to think the issue was with the projectDir variable. I'm fine with either way but this way we wouldn't really suspect if there's any issue with the variables.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -87,9 +87,8 @@ task updateDlsSubmodule(type: Exec) {
 
 compileThrift {
     dependsOn updateDlsSubmodule
-    sourceItems "${projectDir}/src/main/idls/thrift/cadence.thrift","${projectDir}/src/main/idls/thrift/shared.thrift","${projectDir}/src/main/idls/thrift/shadower.thrift"
-
-    nowarn true
+    verbose true
+    sourceItems "./src/main/idls/thrift/cadence.thrift","./src/main/idls/thrift/shared.thrift","./src/main/idls/thrift/shadower.thrift"
 }
 
 compileJava {
--- a/build.gradle
+++ b/build.gradle
@@ -88,7 +88,7 @@ task updateDlsSubmodule(type: Exec) {
 compileThrift {
     dependsOn updateDlsSubmodule
     verbose true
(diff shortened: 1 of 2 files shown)


docker/buildkite/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM adoptopenjdk/openjdk11:alpine
+FROM adoptopenjdk/openjdk11@sha256:13475f4be7d6701a2ebd0516c4c929f2fa78ea773b4483bb580a36a04975e9f9
```
longquanzheng, line 1: Nit: since people like me don’t know the context, would you mind leaving some comments about why we have to pin the sha revision ?

longquanzheng (approved):
Lgtm. I watched how you fixed it it seemed very painful. Is there any idea to prevent or improve in the future? If we have some nightly run integ test would it be easier to discover and fix earlier?

build.gradle
```diff
@@ -87,9 +87,8 @@ task updateDlsSubmodule(type: Exec) {
 
 compileThrift {
     dependsOn updateDlsSubmodule
-    sourceItems "${projectDir}/src/main/idls/thrift/cadence.thrift","${projectDir}/src/main/idls/thrift/shared.thrift","${projectDir}/src/main/idls/thrift/shadower.thrift"
-
-    nowarn true
+    verbose true
+    sourceItems "./src/main/idls/thrift/cadence.thrift","./src/main/idls/thrift/shared.thrift","./src/main/idls/thrift/shadower.thrift"
```
emrahs, line 91: Why do we need to remove ${projectDir}?

demirkayaender: Discussed offline putting this back

demirkayaender:
> Is there any idea to prevent or improve in the future? If we have some nightly run integ test would it be easier to discover and fix earlier?

We have continuous builds, I'm thinking we should create alerts on it so we can get notified when it breaks. Thanks everyone for the quick review! I will address the comments

PR #626, opened 2021-07-13
Revert "Java 11 support (#609)" to fix build

demirkayaender:
Sending just to try if this was the cause for the build failures.


demirkayaender:
It's fixed here: https://github.com/uber/cadence-java-client/pull/625

PR #628, opened 2021-08-07
Add AuthorizationProvider to ClientOptions and WorkflowService

longquanzheng:
For https://docs.google.com/document/d/1a-Xo9FSCLubYlIQHB_eX5j4Yv2GDv_82J94At2dAL3o/edit#

Tested with https://github.com/uber/cadence-java-samples/pull/55 and local auth server
```
./cadence-server --env development_oauth start
```


src/main/java/com/uber/cadence/serviceclient/auth/AdminJwtAuthorizationProvider.java
```diff
@@ -0,0 +1,49 @@
...
+
+  public AdminJwtAuthorizationProvider(RSAPublicKey publicKey, RSAPrivateKey privateKey) {
+    this.rsaPrivateKey = privateKey;
+    this.rsaPublicKey = publicKey;
+  }
+
+  @Override
+  public byte[] getAuthToken() {
+    final JWTCreator.Builder jwtBuilder = JWT.create();
+    jwtBuilder.withClaim("admin", true);
```
mkolodezny, line 42: nit: maybe move the claim mapping into a separate interface and have the admin as a default one

src/main/java/com/uber/cadence/serviceclient/WorkflowServiceTChannel.java
```diff
@@ -264,7 +265,14 @@ private static String getEndpoint(String service, String method) {
         new ThriftRequest.Builder<>(options.getServiceName(), endpoint);
     // Create a mutable hashmap for headers, as tchannel.tracing.PrefixedHeadersCarrier assumes
     // that it can call put directly to add new stuffs (e.g. traces).
-    builder.setHeaders(new HashMap<>(thriftHeaders));
+    final HashMap<String, String> headers = new HashMap<>(thriftHeaders);
+    if (this.options.getAuthProvider() != null) {
```
mkolodezny, line 269: I believe once we have grpc that needs to be extended with TLS case as Temporal did: https://github.com/temporalio/sdk-go/blob/master/internal/grpc_dialer.go#L73

src/main/java/com/uber/cadence/serviceclient/auth/AdminJwtAuthorizationProvider.java
```diff
@@ -0,0 +1,49 @@
...
+    jwtBuilder.withClaim("ttl", 60 * 10);
```
mkolodezny, line 43: nit: we should think about different roles, but yes, for now lets keep it simple

mkolodezny (approved):
The overall looks good to me, assuming it is simplified version of https://docs.temporal.io/docs/server/security/

src/main/java/com/uber/cadence/serviceclient/WorkflowServiceTChannel.java
mkolodezny: cc @vytautas-karpavicius

mkolodezny (commented):
how we are going to handle/extend it to user groups?

src/main/java/com/uber/cadence/serviceclient/auth/AdminJwtAuthorizationProvider.java
longquanzheng: This is a sample provider(also for testing). In the production scenario, they should only set "write" permission/role. The private key should be managed elsewhere should it's impossible to claim a "admin" role here. 

Yeah we could add more roles later, like worker also makes sense to be separate from "write". 
So that
read>write>worker>admin 

because PollForDecision API shouldn't be called from regular user who wants to start/terminate workflows usually.

src/main/java/com/uber/cadence/serviceclient/WorkflowServiceTChannel.java
longquanzheng: Yes, it should be extendable to use header in gRPC.

longquanzheng:
> how we are going to handle/extend it to user groups?

This PR only defines the interface and provides a testing impl. 

For Uber, you need to implement this interface internally. 
The internal implementation will talk to an Auth service(needs to build it) which holds the private key to authorize and return JWT. Then do some caching locally in memory. I can walk you through details in chat.

longquanzheng:
> The overall looks good to me, assuming it is simplified version of https://docs.temporal.io/docs/server/security/

In most parts yes, but some parts this is more advanced like I mentioned in Slack :)

src/main/java/com/uber/cadence/serviceclient/auth/AdminJwtAuthorizationProvider.java
longquanzheng: I don't quite follow this. Are you referring to JWT Mapping interface like the one in the Temporal server? 
I don't think that's applicable here, as this is a very simple implementation for testing.

longquanzheng:
merged in https://github.com/uber/cadence-java-client/pull/632

PR #631, opened 2021-08-24
fix(sticky): checking get workflow history when sticky decision task

yux0:
We see NPE in ReplayDecider with line: decisionTask.getHistory().getEvents().get(0).getWorkflowExecutionStartedEventAttributes();

In ReplayDecisionTaskHandler, we have special handling for sticky decision task that has a partial history. But we do not check the returned history. So add a check to get workflow history when sticky decision task


shijiesheng:
close ancient PR

PR #637, opened 2021-08-31
Fix metrics in WorkflowServiceTChannel

### LINKED ISSUE #636: Metrics from WorkflowServiceTChannel aren't published in v2.x
Metrics from WorkflowServiceTChannel aren't published because it's defaulting to NoopScope.

Using cadence-java-client v2.7.8

sokada1221:
fixes #636


sokada1221:
Not needed as v3.x is compatible with older cadence servers.

PR #641, opened 2021-09-29
Add memo and search attributes to child workflow

spmistry:


src/main/java/com/uber/cadence/internal/sync/SyncDecisionContext.java
```diff
@@ -412,6 +412,8 @@ private ExecuteLocalActivityParameters constructExecuteLocalActivityParameters(
             .setWorkflowIdReusePolicy(options.getWorkflowIdReusePolicy())
             .setRetryParameters(retryParameters)
             .setCronSchedule(options.getCronSchedule())
+            .setMemo(options.getMemo())
```
longquanzheng, line 415: look at line 387. I think we also need to do the same thing

longquanzheng:
Also need a few line code to wire up the StartChildWorkflowExecutionParameters and StartChildWorkflowExecutionDecisionAttributes  https://github.com/uber/cadence-java-client/blob/master/src/main/java/com/uber/cadence/internal/replay/WorkflowDecisionContext.java#L156

src/main/java/com/uber/cadence/internal/sync/SyncDecisionContext.java
spmistry: updated

src/main/java/com/uber/cadence/internal/replay/WorkflowDecisionContext.java
```diff
@@ -153,6 +155,20 @@ public void accept(Exception cause) {
...
+    Map<String, byte[]> memoMap = parameters.getMemo();
+    if (memoMap != null) {
+      memoMap.forEach((key, value) -> memo.putToFields(key, ByteBuffer.wrap(value)));
+      attributes.setMemo(memo);
+    }
+
+    SearchAttributes searchAttributes = new SearchAttributes();
+    Map<String, byte[]> searchAttributesMap = parameters.getSearchAttributes();
+    if (searchAttributesMap != null) {
+      searchAttributesMap.forEach((key, value) -> memo.putToFields(key, ByteBuffer.wrap(value)));
```
longquanzheng, line 168: typo: `memo`->`searchAttributes`

spmistry: fixed

longquanzheng (approved):

PR #644, opened 2021-10-01
Fix consistent query interface which caused overloading ambiguity with variable argument

longquanzheng:
For https://github.com/uber/cadence/issues/4526

### DIFF
--- a/src/main/java/com/uber/cadence/client/WorkflowStub.java
+++ b/src/main/java/com/uber/cadence/client/WorkflowStub.java
@@ -17,7 +17,6 @@
 
 package com.uber.cadence.client;
 
-import com.uber.cadence.QueryConsistencyLevel;
 import com.uber.cadence.QueryRejectCondition;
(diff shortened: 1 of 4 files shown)


src/main/java/com/uber/cadence/client/WorkflowStub.java
```diff
@@ -156,27 +155,38 @@
...
 
-  <R> R query(
+  <R> R queryWithOptions(
       String queryType,
-      Class<R> resultClass,
+      QueryOptions options,
       Type resultType,
-      QueryRejectCondition queryRejectCondition,
-      QueryConsistencyLevel queryConsistencyLevel,
+      Class<R> resultClass,
```
longquanzheng, line 189: to reviewer: I put `Class<R> resultClass` right before `Object... args` so that there is no ambiguity

vytautas-karpavicius (approved):
LGTM.

longquanzheng:
Thanks @vytautas-karpavicius

Issue #620, opened 2021-07-01
Implement CadenceChangeVersion


longquanzheng:
https://github.com/uber/cadence/blob/70f3f582c91f1bb2e265287d91390a1d31e2c2de/common/definition/indexedKeys.go#L48
It's a system defined attribute in server
and it's being used by go client by system:
https://github.com/uber-go/cadence-client/blob/90498dff0c292bf08a868b979d867d7b74196079/internal/workflow.go#L990

https://github.com/uber-go/cadence-client/blob/90498dff0c292bf08a868b979d867d7b74196079/internal/internal_task_handlers.go#L1203

We should implement this also in java client similar as go client https://github.com/uber-go/cadence-client/blob/90498dff0c292bf08a868b979d867d7b74196079/internal/internal_task_handlers.go#L1203

longquanzheng:
key PRs for referrence:
https://github.com/uber-go/cadence-client/pull/848

https://github.com/uber-go/cadence-client/pull/889

PR #650, opened 2021-11-01
Add CadenceChangeVersion support

longquanzheng:
For https://github.com/uber/cadence-java-client/issues/620

The PR is implementing  the combination of 
https://github.com/uber-go/cadence-client/pull/848/files
https://github.com/uber-go/cadence-client/pull/889/files

Some difference:

* In GoSDK we try not to use GenerateSequenceID because this will mess up the sequenceID which is used for making the decision map keys.
JavaSDK doesn’t use SequenceID for search attributes. Instead the mapping is based on nextDecisionEventID. I think we don’t need to do anything? LMK if I am wrong

* JavaSDK doesn't have the same logic of `matchReplayWithHistory` so I don't need to do anything to match https://github.com/uber-go/cadence-client/pull/889 

* Instead JavaSDK is using eventID as the key of the decision map. Additional logic is added to skip the eventID increasing during replay and when GetVersion is removed from workflow code(similar to what we did for GetVersion marker)

Added new test(there are already pretty good tests) and tested locally: https://github.com/uber/cadence-java-samples/pull/59 to verify adding and removing GetVersion won't be a problem.


mkolodezny (approved):

mkolodezny (commented):
Thanks much @longquanzheng for the contribution!

PR #663, opened 2022-01-05
Add capacity handler to release polling lock

spmistry:


src/main/java/com/uber/cadence/internal/worker/Poller.java
```diff
@@ -272,7 +272,12 @@ public void run() throws Exception {
         }
         taskExecutor.process(task);
       } finally {
-        pollSemaphore.release();
+        while (true) {
+          if (taskExecutor.hasCapacity()) {
+            pollSemaphore.release();
+            break;
+          }
```
longquanzheng, line 279: can we add an else branch here to sleep for 100ms? Otherwise this will become a busy loop for a long time. 

Also need to wire up  pollerOptions.pollOnlyIfHasCapacity

src/main/java/com/uber/cadence/internal/worker/PollerOptions.java
```diff
@@ -133,6 +138,11 @@ public Builder setPollThreadNamePrefix(String pollThreadNamePrefix) {
       return this;
     }
 
+    public Builder setPollOnlyIfHasCapacity(boolean pollOnlyIfHasCapacity) {
```
longquanzheng, line 141: let's add a comment here to describe what it does, e.g.
```
the poller will check task executor's remaining capacity before polling tasks. This is to prevent task to get started but not being able to execute in time. 
```

src/main/java/com/uber/cadence/internal/worker/PollerOptions.java
```diff
@@ -60,6 +62,8 @@ public static PollerOptions getDefaultInstance() {
 
     private String pollThreadNamePrefix;
 
+    private Boolean pollOnlyIfHasCapacity = Boolean.FALSE;
```
longquanzheng, line 65: How about change to `pollOnlyIfExecutorHasCapacity` to be more precise?

src/main/java/com/uber/cadence/internal/worker/PollerOptions.java
spmistry: added comment

src/main/java/com/uber/cadence/internal/worker/Poller.java
spmistry: fixed

src/main/java/com/uber/cadence/internal/worker/Poller.java
```diff
@@ -256,6 +256,7 @@ public void run() {
   }
 
   private class PollExecutionTask implements Poller.ThrowingRunnable {
+    private final int LOCK_CHECK_SLEEP_MS = 100;
```
longquanzheng, line 259: nit: how about `EXECUTOR_CAPACITY_CHECK_INTERVAL_MS`?

spmistry: updated

src/main/java/com/uber/cadence/internal/worker/PollerOptions.java
spmistry: updated

src/main/java/com/uber/cadence/internal/worker/PollTaskExecutor.java
```diff
@@ -81,6 +81,11 @@ public void process(T task) {
         });
   }
 
+  @Override
+  public boolean hasCapacity() {
+    return taskExecutor.getQueue().remainingCapacity() > 0;
```
longquanzheng, line 86: I think this need to be `taskExecutor.getActiveCount() < taskExecutor.getPoolSize()`

PR #664, opened 2022-01-10
Introduce experimental flag and add logic for capacity based polling

spmistry:
Draft PR Ref: #663 (Closed due to CLA signing issue)


longquanzheng (approved):
Lgtm

PR #673, opened 2022-06-16
add gRPC API adapter

mkolodezny:


src/main/java/com/uber/cadence/internal/compatibility/Thrift2ProtoAdapter.java
```diff
@@ -0,0 +1,899 @@
...
+                RequestMapper.listTaskListPartitionsRequest(request));
+    return ResponseMapper.listTaskListPartitionsResponse(response);
+  }
+
+  @Override
+  public void RefreshWorkflowTasks(RefreshWorkflowTasksRequest request)
+      throws BadRequestError, DomainNotActiveError, ServiceBusyError, EntityNotExistsError,
+      TException {
+    throw new UnsupportedOperationException("not implemented");
+  }
```
just-at-uber, line 570: I assume this is going to implemented in a follow up PR?

mkolodezny: yes, the most of the methods are not used at all so we will follow up as we did for Go SDK

just-at-uber (approved):

PR #683, opened 2022-07-25
do not return null for empty proto collections

mkolodezny:


src/main/java/com/uber/cadence/internal/compatibility/Thrift2ProtoAdapter.java
```diff
@@ -345,6 +345,7 @@ public void RequestCancelWorkflowExecution(RequestCancelWorkflowExecutionRequest
       throws BadRequestError, EntityNotExistsError, CancellationAlreadyRequestedError,
           ServiceBusyError, DomainNotActiveError, LimitExceededError,
           ClientVersionNotSupportedError, WorkflowExecutionAlreadyCompletedError, TException {
+    cancelRequest.setRequestId(UUID.randomUUID().toString());
```
Shaddoll, line 348: I think this shouldn't be done in the converter. If we convert it back, the data won't be equal

mkolodezny: good point, I following the existing pattern for thrift implementation to be safe and fully compatible https://github.com/uber/cadence-java-client/blob/cb5d6ecaeb206044213d9ab2cf45bd74d6f8b654/src/main/java/com/uber/cadence/serviceclient/WorkflowServiceTChannel.java#L2251 
there is the refactoring room afterwards

Shaddoll: Yeah, the UUIDs should be set in the thrift client

mkolodezny: Its a completely different (gRPC) implementation of IWorkflowService with thrift->proto mappers inside to avoid breaking changes for now.

Shaddoll (approved):

PR #731, opened 2022-08-04
add gRPC channel to client options

mkolodezny:


src/main/java/com/uber/cadence/internal/compatibility/proto/serviceclient/GrpcServiceStubs.java
```diff
@@ -80,12 +81,17 @@ final class GrpcServiceStubs implements IGrpcServiceStubs {
...
-            .usePlaintext()
-            .build();
-
+    if (options.getGRPCChannel() != null) {
+      this.channel = options.getGRPCChannel();
+      shutdownChannel = false;
+    } else {
+      this.channel =
+          ManagedChannelBuilder.forAddress(options.getHost(), options.getPort())
+              .defaultLoadBalancingPolicy("round_robin")
```
davidporter-id-au, line 90: is `round_robbin` a constant somewhere?  Should it be?

davidporter-id-au (approved):

src/main/java/com/uber/cadence/internal/compatibility/proto/serviceclient/GrpcServiceStubs.java
mkolodezny: good point, no strong opinion here but prefer go with inline mostly

PR #744, opened 2022-09-12
emitting java client version that is used for a specific domain.

mindaugasbarcauskas:


src/main/java/com/uber/cadence/internal/metrics/ClientVersionEmitter.java
```diff
@@ -0,0 +1,65 @@
...
+    } else {
+      this.metricScope = metricScope;
+    }
+
+    if (domain == null) {
+      domain = "";
+    }
+
+    Properties prop = new Properties();
+    InputStream in = Version.class.getResourceAsStream("/com/uber/cadence/version.properties");
```
davidporter-id-au, line 45: What's the process whereby we update this?

mindaugasbarcauskas: It is generated by gradle, so no need to update manually

src/main/java/com/uber/cadence/internal/sync/WorkflowClientInternal.java
```diff
@@ -42,7 +43,7 @@
 import java.util.Arrays;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.*;
```
vytautas-karpavicius, line 46: nit: please use explicit imports

vytautas-karpavicius (approved):

PR #749, opened 2022-09-21
Supporting graceful shutdown based on SIGTERM handler.

mindaugasbarcauskas:
Adding handler for graceful shutdown. Since uDeploy sends a signal 10 seconds before turning off the service. We want to avoid abruptly stopping any activity or workflow executions.

### DIFF
--- a/src/main/java/com/uber/cadence/internal/sync/WorkflowClientInternal.java
+++ b/src/main/java/com/uber/cadence/internal/sync/WorkflowClientInternal.java
@@ -33,7 +33,9 @@ import com.uber.cadence.internal.external.ManualActivityCompletionClientFactory;
 import com.uber.cadence.internal.external.ManualActivityCompletionClientFactoryImpl;
 import com.uber.cadence.internal.metrics.ClientVersionEmitter;
 import com.uber.cadence.internal.sync.WorkflowInvocationHandler.InvocationType;
+import com.uber.cadence.internal.worker.WorkerShutDownHandler;
 import com.uber.cadence.serviceclient.IWorkflowService;
+import com.uber.cadence.worker.WorkerFactory;
 import com.uber.cadence.workflow.Functions;
 import com.uber.cadence.workflow.QueryMethod;
 import com.uber.cadence.workflow.WorkflowMethod;
@@ -71,6 +73,8 @@ public final class WorkflowClientInternal implements WorkflowClient {
     Objects.requireNonNull(options);
 
     emitClientVersion(options);
+    WorkerShutDownHandler.registerHandler();
+
     return new WorkflowClientInternal(service, options);
   }
 
--- a/src/main/java/com/uber/cadence/internal/sync/WorkflowClientInternal.java
+++ b/src/main/java/com/uber/cadence/internal/sync/WorkflowClientInternal.java
@@ -71,7 +71,6 @@ public final class WorkflowClientInternal implements WorkflowClient {
     Objects.requireNonNull(options);
 
     emitClientVersion(options);
-    WorkerShutDownHandler.registerHandler();
 
     return new WorkflowClientInternal(service, options);
   }
--- a/src/main/java/com/uber/cadence/internal/sync/WorkflowClientInternal.java
+++ b/src/main/java/com/uber/cadence/internal/sync/WorkflowClientInternal.java
@@ -35,7 +35,6 @@ import com.uber.cadence.internal.metrics.ClientVersionEmitter;
 import com.uber.cadence.internal.sync.WorkflowInvocationHandler.InvocationType;
 import com.uber.cadence.internal.worker.WorkerShutDownHandler;
 import com.uber.cadence.serviceclient.IWorkflowService;
-import com.uber.cadence.worker.WorkerFactory;
 import com.uber.cadence.workflow.Functions;
 import com.uber.cadence.workflow.QueryMethod;
 import com.uber.cadence.workflow.WorkflowMethod;
@@ -48,7 +47,6 @@ import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
-
 import org.apache.thrift.TException;
 
 public final class WorkflowClientInternal implements WorkflowClient {
@@ -412,7 +410,7 @@ public final class WorkflowClientInternal implements WorkflowClient {
     return execute(() -> workflow.apply(arg1, arg2, arg3, arg4, arg5, arg6));
   }
 
-  private synchronized static void emitClientVersion(WorkflowClientOptions options) {
+  private static synchronized void emitClientVersion(WorkflowClientOptions options) {
     if (emittingClientVersion) {
       return;
     }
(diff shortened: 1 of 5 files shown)


src/main/java/com/uber/cadence/internal/worker/WorkerShutDownHandler.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+public class WorkerShutDownHandler {
+
+    private static final List<WorkerFactory> workerFactories = new ArrayList<>();
+    private static boolean registered = false;
+
+    public static void registerHandler() {
+        if (!registered) {
```
vytautas-karpavicius, line 15: nit: invert if to return early

src/main/java/com/uber/cadence/internal/worker/WorkerShutDownHandler.java
```diff
@@ -0,0 +1,42 @@
...
+                    }
+                }
+            });
+        }
+    }
+
+
+    public static void registerWorkerFactory(WorkerFactory workerFactory) {
+        if (workerFactory != null) {
+            workerFactories.add(workerFactory);
```
vytautas-karpavicius, line 39: Is it possible for this to be accessed from several threads? I think it may be a good idea to synchronize this method.

src/main/java/com/uber/cadence/internal/worker/WorkerShutDownHandler.java
```diff
@@ -0,0 +1,61 @@
...
+                for (WorkerFactory workerFactory : workerFactories) {
+                  workerFactory.suspendPolling();
+                }
+
+                for (WorkerFactory workerFactory : workerFactories) {
+                  workerFactory.shutdownNow();
+                }
+
+                for (WorkerFactory workerFactory : workerFactories) {
+                  workerFactory.awaitTermination(1, TimeUnit.SECONDS);
```
emrahs, line 50: 1 second per workerFactory seems arbitrary without knowing how many of them exists. I'd consider passing the total available time (ie. 10 seconds) to the function and return the remaining time that can be used for the next workerFactory. You can see how that pattern works inside the `awaitTermination`, so you can basically repeat the same pattern here, too.

src/main/java/com/uber/cadence/internal/sync/WorkflowClientInternal.java
```diff
@@ -71,6 +71,8 @@ public static WorkflowClient newInstance(
     Objects.requireNonNull(options);
 
     emitClientVersion(options);
+    WorkerShutDownHandler.registerHandler();
```
emrahs, line 74: This seems like a weird place to register worker shutdown handler, because the WorkflowClient can also be used for non-worker purposes (ie. starting/signaling a workflow from any Java service). 

How about doing the registration in the WorkerFactory, ie. WorkerFactor.newInstance()?

src/main/java/com/uber/cadence/worker/WorkerFactory.java
```diff
@@ -136,6 +137,8 @@ public WorkerFactory(WorkflowClient workflowClient, WorkerFactoryOptions factory
                 .setPollThreadCount(this.factoryOptions.getStickyPollerCount())
                 .build(),
             stickyScope);
+
+    WorkerShutDownHandler.registerWorkerFactory(this);
```
emrahs, line 141: Adding some tests would be great.

src/main/java/com/uber/cadence/internal/worker/WorkerShutDownHandler.java
```diff
@@ -0,0 +1,70 @@
+/*
+ *  Modifications Copyright (c) 2017-2020 Uber Technologies Inc.
+ *  Portions of the Software are attributed to Copyright (c) 2020 Temporal Technologies Inc.
```
emrahs, line 3: Since this is a brand new file that we are adding without being inspired by Temporal, you can remove this copyright note.

Please also change 2020 to 2022.

src/main/java/com/uber/cadence/internal/worker/WorkerShutDownHandler.java
```diff
@@ -0,0 +1,70 @@
...
+
+        for (WorkerFactory workerFactory : workerFactories) {
+          workerFactory.shutdownNow();
+        }
+
+        long remainingTimeout = 10000;
+        for (WorkerFactory workerFactory : workerFactories) {
+          final long timeoutMillis = remainingTimeout;
+          remainingTimeout = InternalUtils.awaitTermination(timeoutMillis,
+                  () -> workerFactory.awaitTermination(timeoutMillis, TimeUnit.MILLISECONDS));
```
emrahs, line 51: nit: you can specify this in seconds by using TimeUnit.Seconds

src/test/java/com/uber/cadence/internal/worker/WorkerShutDownHandlerTest.java
```diff
@@ -0,0 +1,68 @@
...
+    public void setup() {
+        WorkflowClientOptions clientOptions =
+                WorkflowClientOptions.newBuilder().setMetricsScope(new NoopScope()).build();
+        when(mockClient.getOptions()).thenReturn(clientOptions);
+        when(mockClient.getService()).thenReturn(mockService);
+    }
+
+    @Test
+    public void shutDownHookShutsDownFactories() {
+        WorkerShutDownHandler.registerHandler();
```
emrahs, line 53: WorkerFactor.newInstance already takes care of registration, so you can remove this. In fact, it's much better to remove this so that the test can fail if someone removes the handler registration unintentionally. 

With that, the test is mostly ensuring that the shutdown handler is doing its job. The second thing we need to check is whether the shutdown handler is actually triggered by JVM (ie. registration is done right). I think this isn't a fully unit-testable thing, so perhaps you can at least check it manually with Ctrl + C and some logging.

emrahs (commented):
Thanks for all the changes! I only left a few minor comments.

src/main/java/com/uber/cadence/internal/worker/WorkerShutDownHandler.java
mindaugasbarcauskas: I prefer to keep this as milliseconds since InternalUtils.awaitTermination calculated the remainingTimeout in milliseconds.

src/test/java/com/uber/cadence/internal/worker/WorkerShutDownHandlerTest.java
mindaugasbarcauskas: good point! Tested manually and the handler is executed before shutdown.

src/main/java/com/uber/cadence/internal/sync/WorkflowClientInternal.java
```diff
@@ -33,6 +33,7 @@
 import com.uber.cadence.internal.external.ManualActivityCompletionClientFactoryImpl;
 import com.uber.cadence.internal.metrics.ClientVersionEmitter;
 import com.uber.cadence.internal.sync.WorkflowInvocationHandler.InvocationType;
+import com.uber.cadence.internal.worker.WorkerShutDownHandler;
```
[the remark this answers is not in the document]
mindaugasbarcauskas: removed.

src/main/java/com/uber/cadence/internal/worker/WorkerShutDownHandler.java
mindaugasbarcauskas: done

vytautas-karpavicius (approved):

emrahs (approved):

PR #754, opened 2022-10-11
Adding cross domain signaling.

mindaugasbarcauskas:
Example use:

WorkflowExecution workflowExecution = new WorkflowExecution().setWorkflowId(workflowId);
      ExternalWorkflowStub workflowStub =
          Workflow.newUntypedExternalWorkflowStub(workflowExecution);
      workflowStub.signalCrossDomain("FareWellWorkflow::waitForName", "testdomain2", "Mindaugas");

### DIFF
--- a/src/main/java/com/uber/cadence/internal/sync/ExternalWorkflowStubImpl.java
+++ b/src/main/java/com/uber/cadence/internal/sync/ExternalWorkflowStubImpl.java
@@ -22,6 +22,7 @@ import com.uber.cadence.workflow.CancelExternalWorkflowException;
 import com.uber.cadence.workflow.ExternalWorkflowStub;
 import com.uber.cadence.workflow.Promise;
 import com.uber.cadence.workflow.SignalExternalWorkflowException;
+import com.uber.cadence.workflow.SignalOptions;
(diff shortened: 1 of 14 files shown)


src/main/java/com/uber/cadence/internal/sync/ExternalWorkflowStubImpl.java
```diff
@@ -59,6 +59,24 @@ public void signal(String signalName, Object... args) {
     }
   }
 
+  @Override
+  public void signalCrossDomain(String signalName, String domain, Object... args) {
```
mindaugasbarcauskas, line 63: Probably similar change is needed in ChildWorkflowStub?

mindaugasbarcauskas: @Groxx maybe you know?

davidporter-id-au: I believe that's the case

davidporter-id-au (approved):
LGTM inasmuch as I'm pretty poor at reading Java, thank you

PR #757, opened 2022-10-24
Adding open tracing support.

mindaugasbarcauskas:
Adding support for opentracing. Now if user has a span defined then they will see calls going to cadence server side from the client.

example view:

image


src/main/java/com/uber/cadence/internal/compatibility/proto/serviceclient/GrpcServiceStubs.java
```diff
@@ -103,7 +107,10 @@ final class GrpcServiceStubs implements IGrpcServiceStubs {
     headers.put(RPC_ENCODING_HEADER_KEY, "proto");
     Channel interceptedChannel =
         ClientInterceptors.intercept(
-            channel, deadlineInterceptor, MetadataUtils.newAttachHeadersInterceptor(headers));
+            channel,
+            deadlineInterceptor,
+            MetadataUtils.newAttachHeadersInterceptor(headers),
+            newOpenTelemetryInterceptor());
```
mantas-sidlauskas, line 113: is this going to be created per request basis? or a single interceptor will be reused?

mindaugasbarcauskas: This interceptor is used for interceptedChannel instance creation which is created per GrcpServiceStubs instance.

mantas-sidlauskas (approved):

PR #760, opened 2022-10-27
Adding client option to override shutdown timeout.

mindaugasbarcauskas:
Adding option so that it is possible to override the time allotted for worker shutdown after SIGTERM signal.


shijiesheng:
What error are we seeing without this change? Closed it for now.

PR #761, opened 2022-11-10
Adding poller autoscaler, to dynamically resize number of pollers.

mindaugasbarcauskas:
Initial implementation for poller autoscaler.

How it works:
Feature will be under feature flag.
If enabled, every 60 seconds it will recalculate the optimal number of pollers based on the amount of empty polls vs polls that return tasks. 

the pollerThreadCount is treated as the defined maximum so autoscaler can only scale between 1poller and the defined poller amount.

Resizable semaphore is used for controlling number of polls allowed.

### DIFF
--- a/src/main/java/com/uber/cadence/internal/worker/autoscaler/PollerAutoScaler.java
+++ b/src/main/java/com/uber/cadence/internal/worker/autoscaler/PollerAutoScaler.java
@@ -24,12 +24,12 @@ public class PollerAutoScaler {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(PollerAutoScaler.class);
 
-  private Duration coolDownTime;
-  private PollerUsageEstimator pollerUsageEstimator;
-  private Recommender recommender;
-  private ResizableSemaphore semaphore;
+  private final Duration coolDownTime;
+  private final PollerUsageEstimator pollerUsageEstimator;
+  private final Recommender recommender;
+  private final ResizableSemaphore semaphore;
   private int semaphoreSize;
-  private boolean shutingDown;
+  private boolean shuttingDown;
 
   public PollerAutoScaler(
       Duration coolDownTime, PollerUsageEstimator pollerUsageEstimator, Recommender recommender) {
@@ -46,13 +46,14 @@ public class PollerAutoScaler {
             new Runnable() {
               @Override
               public void run() {
-                while (!shutingDown) {
+                while (!shuttingDown) {
                   try {
                     Thread.sleep(coolDownTime.toMillis());
-                    if (!shutingDown) {
+                    if (!shuttingDown) {
                       resizePollers();
                     }
                   } catch (InterruptedException e) {
+                    LOGGER.info("interrupted wait for next poller scaling");
                   }
                 }
               }
@@ -61,7 +62,7 @@ public class PollerAutoScaler {
 
   public void stop() {
     LOGGER.info("shutting down poller autoscaler");
-    shutingDown = true;
+    shuttingDown = true;
   }
 
   protected void resizePollers() {
@@ -86,14 +87,6 @@ public class PollerAutoScaler {
 
   public int getUpperPollerAmount() {
     return recommender.getUpperValue();
-  }
-
-  public PollerUsageEstimator getPollerUsageEstimator() {
-    return pollerUsageEstimator;
-  }
-
-  public Recommender getRecommender() {
-    return recommender;
   }
 
   public void acquire() throws InterruptedException {
--- a/src/main/java/com/uber/cadence/internal/worker/autoscaler/PollerAutoScaler.java
+++ b/src/main/java/com/uber/cadence/internal/worker/autoscaler/PollerAutoScaler.java
@@ -1,0 +1,111 @@
+/*
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"). You may not
+ *  use this file except in compliance with the License. A copy of the License is
+ *  located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package com.uber.cadence.internal.worker.autoscaler;
+
+import java.time.Duration;
+import java.util.concurrent.Executors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PollerAutoScaler {
+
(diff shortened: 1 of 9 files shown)


src/main/java/com/uber/cadence/internal/worker/autoscaler/Recommender.java
```diff
@@ -0,0 +1,50 @@
...
+
+  public int recommend(int currentPollers, float pollerUtilizationRate) {
+    int recommended = 0;
+
+    if (pollerUtilizationRate == 1) {
+      return upperValue;
+    }
+
+    float r = currentPollers * pollerUtilizationRate / targetPollerUtilRate;
+    r = Math.min(upperValue, Math.max(lowerValue, r));
```
shijiesheng, line 38: nit: `return Math.min(upperValue, Math.max(lowerValue, r));` might be simpler?

src/main/java/com/uber/cadence/internal/worker/autoscaler/PollerAutoScaler.java
```diff
@@ -0,0 +1,111 @@
...
+  public int getUpperPollerAmount() {
+    return recommender.getUpperValue();
+  }
+
+  public PollerUsageEstimator getPollerUsageEstimator() {
+    return pollerUsageEstimator;
+  }
+
+  public Recommender getRecommender() {
+    return recommender;
```
shijiesheng, line 96: all recommender apis are exposed above, do you still need to get this entity?

src/main/java/com/uber/cadence/internal/worker/autoscaler/PollerAutoScaler.java
```diff
@@ -0,0 +1,111 @@
...
+            new Runnable() {
+              @Override
+              public void run() {
+                while (!shutingDown) {
+                  try {
+                    Thread.sleep(coolDownTime.toMillis());
+                    if (!shutingDown) {
+                      resizePollers();
+                    }
+                  } catch (InterruptedException e) {
```
shijiesheng, line 55: do you need to handle anything here (maybe logging)

src/main/java/com/uber/cadence/internal/worker/autoscaler/PollerUsageEstimator.java
```diff
@@ -0,0 +1,45 @@
...
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package com.uber.cadence.internal.worker.autoscaler;
+
+public class PollerUsageEstimator {
+
+  private int noopTaskCount;
```
shijiesheng, line 20: This is not threadsafe implementation. There will be concurrencies to tally the actionable and noop tasks. You'll need either locks or atomic variables.

src/test/java/com/uber/cadence/internal/worker/autoscaler/PollerUsageEstimatorTest.java
```diff
@@ -0,0 +1,51 @@
+package com.uber.cadence.internal.worker.autoscaler;
+
+import static org.junit.Assert.*;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.runners.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class PollerUsageEstimatorTest {
```
shijiesheng, line 10: you might want to add concurrencies in the test to ensure thread safety.

src/main/java/com/uber/cadence/internal/worker/autoscaler/Recommender.java
mindaugasbarcauskas: yeah makes sense, will update.

src/main/java/com/uber/cadence/internal/worker/autoscaler/PollerAutoScaler.java
mindaugasbarcauskas: Not really, will review all the interfaces.

src/main/java/com/uber/cadence/internal/worker/autoscaler/PollerAutoScaler.java
mindaugasbarcauskas: Don't think it is necessary, but might add some log just for obervability.

src/main/java/com/uber/cadence/internal/worker/autoscaler/PollerUsageEstimator.java
mindaugasbarcauskas: Good catch.

src/main/java/com/uber/cadence/internal/worker/autoscaler/PollerUsageEstimator.java
```diff
@@ -0,0 +1,48 @@
...
+    int noopTasks = noopTaskCount.get();
+    if (noopTasks + actionableTasks == 0) {
+      return new PollerUsage(0);
+    }
+    PollerUsage result = new PollerUsage((actionableTasks * 1f) / (noopTasks + actionableTasks));
+    reset();
+    return result;
+  }
+
+  public void reset() {
```
vytautas-karpavicius, line 44: Could this be private? I do not see it being used elsewhere.

src/main/java/com/uber/cadence/internal/worker/autoscaler/ResizableSemaphore.java
```diff
@@ -0,0 +1,28 @@
...
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package com.uber.cadence.internal.worker.autoscaler;
+
+import java.util.concurrent.Semaphore;
+
+public class ResizableSemaphore extends Semaphore {
```
vytautas-karpavicius, line 20: What functionality this class adds? Could we just use base class directly?

src/main/java/com/uber/cadence/internal/worker/autoscaler/PollerAutoScaler.java
```diff
@@ -0,0 +1,104 @@
...
+      semaphore.release(diff * -1);
+    } else {
+      semaphore.decreasePermits(diff);
+    }
+
+    LOGGER.info(String.format("resized pollers to: %d", pollerCount));
+    this.semaphoreSize = pollerCount;
+  }
+
+  public int getLowerPollerAmount() {
```
vytautas-karpavicius, line 84: Where is the intended usage of `getLowerPollerAmount` / `getUpperPollerAmount` ?

src/main/java/com/uber/cadence/internal/worker/autoscaler/ResizableSemaphore.java
mindaugasbarcauskas: Subclass is needed because reducePermits is protected method.

PR #762, opened 2022-11-11
Making dynamic poller configurable.

mindaugasbarcauskas:
Starting to use dynamic poller and making it configurable.

Initial dynamic poller implementation:

https://github.com/uber/cadence-java-client/pull/761

**Why dynamic poller is needed?**

Dynamic poller allows lowering the amount of polls for bursty traffic (when there are no workflows we downsize the poller amount to ease the load on the server)

**Testing**

This was tested by running cadence-java-sample

Observing pollers dropping to 1 because there are no tasks available.
Executing 100 workflows 
Seeing pollers increase to handle the available tasks
Then observing again poller amount decreasing since there are no more workflows left to process.

Additionally unit tests are available for the different components like Recommender, UsageEstimator.

### DIFF
--- a/src/main/java/com/uber/cadence/internal/worker/Poller.java
+++ b/src/main/java/com/uber/cadence/internal/worker/Poller.java
@@ -21,15 +21,11 @@ import com.uber.cadence.internal.common.BackoffThrottler;
 import com.uber.cadence.internal.common.InternalUtils;
 import com.uber.cadence.internal.metrics.MetricsType;
 import com.uber.cadence.internal.worker.autoscaler.AutoScaler;
-import com.uber.cadence.internal.worker.autoscaler.NoopAutoScaler;
-import com.uber.cadence.internal.worker.autoscaler.PollerAutoScaler;
-import com.uber.cadence.internal.worker.autoscaler.PollerUsageEstimator;
-import com.uber.cadence.internal.worker.autoscaler.Recommender;
+import com.uber.cadence.internal.worker.autoscaler.AutoScalerFactory;
 import com.uber.m3.tally.Scope;
 import java.util.Objects;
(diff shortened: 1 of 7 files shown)


src/main/java/com/uber/cadence/internal/worker/Poller.java
```diff
@@ -267,14 +291,22 @@ private class PollExecutionTask implements Poller.ThrowingRunnable {
     @Override
     public void run() throws Exception {
       try {
-        pollSemaphore.acquire();
-        T task = pollTask.poll();
-        if (task == null) {
-          return;
+        pollerAutoScaler.acquire();
+        try {
+          pollSemaphore.acquire();
```
vytautas-karpavicius, line 296: Seems like original `pollSemaphore` does not participate in any meaningful way. It is created for every `PollExecutionTask` separately and is not shared. I suggest removing this entirely.

src/main/java/com/uber/cadence/internal/worker/Poller.java
```diff
@@ -86,6 +93,20 @@ public Poller(
     this.taskExecutor = taskExecutor;
     this.pollerOptions = pollerOptions;
     this.metricsScope = metricsScope;
+
+    if (pollerOptions.getPollerAutoScalerOptions() != null) {
```
vytautas-karpavicius, line 97: nit: I suggest to put all building details inside autoscaler package instead.

src/main/java/com/uber/cadence/internal/worker/Poller.java
mindaugasbarcauskas: Agree, removed it.

vytautas-karpavicius (approved):

PR #764, opened 2022-11-18
Large activity parameter support

mindaugasbarcauskas:
Feature for passing around large amounts of data to activities and not hitting workflow history size limits since the data can be stored in external storage.


shijiesheng:
We are now building large blob support and will start in python SDK inside data converter.

PR #772, opened 2022-12-23
Adding ability to override activity options.

mindaugasbarcauskas:
Adding ability to override activity options by passing a map with keys that are activityName and the value is the activityOption.

This will allow to override the activityOptions for different implementations of an activity interface. Please check the accompanying WorkflowTest written to see the functionality. For example:

You have activity interface I and implementations A,B. You may want to have different StartToClose timeouts for A and B and before that was not possible so this feature will allow to do it.

### DIFF
--- a/src/main/java/com/uber/cadence/internal/sync/ActivityInvocationHandler.java
+++ b/src/main/java/com/uber/cadence/internal/sync/ActivityInvocationHandler.java
@@ -24,6 +24,7 @@ import com.uber.cadence.workflow.ActivityStub;
 import com.uber.cadence.workflow.WorkflowInterceptor;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
+import java.util.Map;
 import java.util.function.Function;
 
 class ActivityInvocationHandler extends ActivityInvocationHandlerBase {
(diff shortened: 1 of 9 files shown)


src/main/java/com/uber/cadence/internal/sync/ActivityInvocationHandler.java
```diff
@@ -43,9 +44,30 @@ private ActivityInvocationHandler(ActivityOptions options, WorkflowInterceptor a
...
+    }
+
     Function<Object[], Object> function;
     ActivityOptions mergedOptions = ActivityOptions.merge(activityMethod, methodRetry, options);
-    ActivityStub stub = ActivityStubImpl.newInstance(mergedOptions, activityExecutor);
+    ActivityStub stub;
+    if (activityOptionsOverride == null) {
+      stub = ActivityStubImpl.newInstance(mergedOptions, activityExecutor);
+    } else {
+      stub = ActivityStubImpl.newInstance(activityOptionsOverride, activityExecutor);
```
mindaugasbarcauskas, line 69: Might makes sense to merge override options as well.

src/main/java/com/uber/cadence/worker/WorkflowImplementationOptions.java
```diff
@@ -19,13 +19,17 @@
...
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Objects;
 
 public final class WorkflowImplementationOptions {
 
   public static final class Builder {
 
     private NonDeterministicWorkflowPolicy nonDeterministicWorkflowPolicy = BlockWorkflow;
+    private Map<String, ActivityOptions> activityOptions = new HashMap<>();
```
davidporter-id-au, line 32: might be worth a comment as to what the intent of the string key is here, as I understand it it's the identifier of the activity, but just from looking at the code that's not super obvious

src/main/java/com/uber/cadence/internal/sync/ActivityInvocationHandler.java
davidporter-id-au: yeah, as discussed, only if there's a sense in which the interface options are 'default', if I've understood that correctly. It might be more ergonomic to just override specific options rather than the entire set.

davidporter-id-au (approved):
Weakly I'd consider the override thing a worthwhile add, but otherwise lgtm, I don't know enough about Java's ecosystem to know if this is the idiomatic, but it seems reasonable and easy to use.

src/main/java/com/uber/cadence/worker/WorkflowImplementationOptions.java
mindaugasbarcauskas: there is a comment that says that the key is the activity name. The comment is above setActivityOptionOverrides method. I assume it should be fine?

PR #776, opened 2023-03-16
M1 mac fix, document doNotCompleteOnReturn's behavioral gap explicitly

Groxx:
Java counterpart to https://github.com/uber-go/cadence-client/pull/1229

Resolving this "two kinds of finished-timeouts / heartbeat-timeouts, but only one config" issue with Activity.doNotCompleteOnReturn will need some new semantics on the server side... but at least there's a workaround until then.

The general concept of a "pending activity" does seem worth supporting, particularly because it can respond with an "already completed" error when a second "complete" request is received. It just has unfortunate quirks at the moment.

---

Since I'm on an M1, I've also fixed the gradle project for M1 macs.
Upgrading these dependencies would also likely be good, but is a much larger change than "lean on x86 emulation to do the exact same thing we're already doing".

And as a side note: VSCode was an absolute nightmare to try to use here, I suspect because of a lack of configurability + old versions.  Intellij (community edition) picked everything up automatically though.  It might be worth recommending it in contributing docs.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -119,11 +119,19 @@ sourceSets {
 
 protobuf {
     protoc {
-        artifact = 'com.google.protobuf:protoc:3.11.0'
+        if (osdetector.os == "osx") {
+            artifact = 'com.google.protobuf:protoc:3.11.0:osx-x86_64' // no arm version available
+        } else {
+            artifact = 'com.google.protobuf:protoc:3.11.0'
+        }
     }
     plugins {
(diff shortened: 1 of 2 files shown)


Groxx:
```
      <h1>Down for Maintenance</h1>
      <p>FOSSA is down for maintenance. We&rsquo;ll be back soon.</p>
```
guess I'll retry in a bit.

PR #779, opened 2023-04-13
Add executeWorkflow method for WorkflowInterceptor to match go client library

abhishekj720:
**Background**

Go client supports intercepting execute workflow in the workflow interceptor. This change is to implement the same behavior. 

**Changes**

* WorkflowInterceptor interface change
* WorkflowInterceptorBase change (no-op)
* SyncDecisionContext change (default behavior if no additional interceptor)
* fix test implementations of interceptor that is not extending WorkflowInterceptorBase


src/main/java/com/uber/cadence/workflow/WorkflowInterceptor.java
```diff
@@ -51,6 +51,11 @@ public Promise<WorkflowExecution> getWorkflowExecution() {
     }
   }
 
+  // interceptor executeWorkflow method: no-op interceptors delegate calls to next interceptors used
```
shijiesheng, line 54: This comment should go to WorkflowInterceptorBase. 
Here you might want to stress that this is to match `interceptor:ExecuteWorkflow` behavior in go client.

shijiesheng: Also, you want to use [javadoc tags](https://www.javatpoint.com/java-comments) to illustrate the meaning of input and return value. Otherwise, it's confusing to have byte[] array input as output.

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -6118,6 +6120,19 @@ private TracingWorkflowInterceptor(FilteredTrace trace, WorkflowInterceptor next
...
 
+    /**
+     * @param workflowDefinition
+     * @param input
+     * @return
+     * @throws CancellationException
+     * @throws WorkflowExecutionException
+     */
+    @Override
+    public byte[] executeWorkflow(SyncWorkflowDefinition workflowDefinition, byte[] input)
```
shijiesheng, line 6131: Instead of using raw input, let's create an entity to include WorkflowStartEventAttributes.

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -6118,6 +6120,19 @@ private TracingWorkflowInterceptor(FilteredTrace trace, WorkflowInterceptor next
...
+        throws CancellationException, WorkflowExecutionException {
+      return next.executeWorkflow(workflowDefinition, input);
```
shijiesheng, line 6133: you'll need to add trace here.

src/main/java/com/uber/cadence/workflow/WorkflowInterceptor.java
```diff
@@ -51,6 +51,10 @@ public Promise<WorkflowExecution> getWorkflowExecution() {
     }
   }
 
+  //to match behavior in go client: interceptor executeWorkflow method
+  byte[] executeWorkflow(SyncWorkflowDefinition workflowDefinition, byte[] input)
```
shijiesheng, line 55: instead of raw input, we might want to create a separate entity to include WorkflowStartEventAttributes

src/main/java/com/uber/cadence/workflow/WorkflowInterceptor.java
```diff
@@ -18,20 +18,42 @@
...
-import java.util.Random;
-import java.util.UUID;
+import java.util.*;
+import java.util.concurrent.CancellationException;
 import java.util.function.BiPredicate;
 import java.util.function.Supplier;
 
 public interface WorkflowInterceptor {
 
+  public final class WorkflowAttributeInput {
```
shijiesheng, line 37: `WorkflowInput` might be a better naming

src/test/java/com/uber/cadence/workflow/interceptors/SignalWorkflowInterceptor.java
```diff
@@ -47,6 +50,20 @@ public SignalWorkflowInterceptor(
     this.next = Objects.requireNonNull(next);
   }
 
+  /**
```
shijiesheng, line 53: comments are not needed for test classes

src/main/java/com/uber/cadence/workflow/WorkflowInterceptorBase.java
```diff
@@ -40,6 +38,15 @@ public WorkflowInterceptorBase(WorkflowInterceptor next) {
     this.next = next;
   }
 
+  // base interceptor executeWorkflow method: no-op interceptors delegate calls to next interceptors
+  // used
```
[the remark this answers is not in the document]
abhishekj720: these are auto linting format worker

src/test/java/com/uber/cadence/workflow/interceptors/SignalWorkflowInterceptor.java
Groxx: particularly when they provide less information than the type signature

shijiesheng (approved):

PR #782, opened 2023-04-19
Added hook for Google Java Format enforcement

abhishekj720:
Added pre-commit hook: whenever new code is being pushed, code enforcement will auto format the code.

Once added: please delete any pre-commit file if already exists in .git/hooks and run command (in .git/hooks) : "ln -s ../../hooks/pre-commit pre-commit"  to make a symbolic link pointing to the pre-commit file present in root directory's hooks.

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+# This hook automatically enforces the Google Java Format style on Java files using the goJF Gradle task.
+
+# Change directory to the root of the repository
+cd "$(git rev-parse --show-toplevel)"
+
+# Run the goJF Gradle task on all Java files that have been modified
+./gradlew goJF --quiet $(git diff --cached --name-only --diff-filter=ACM -- '*.java')
+


Groxx:
Covered in chat internally, but as a tl;dr: commit hooks are WAY more pain than they're worth, :-1:.  Every single use I've seen has eventually removed them with enthusiasm due to either significant conflicts with reasonable workflows, or absolutely massive numbers of people simply not using them.

If we're missing CI checks: yea, we should _absolutely_ plug those.
If we're too _difficult_ to run CI checks locally, for earlier warning: yea, we should absolutely improve that.
If we're encountering people not reading instructions, and getting caught by CI, and annoyed at the delay: working as intended.  Hooks require reading instructions too, so they won't help with this.  Reading instructions _must_ be assumed, it's better that we can't run arbitrary code on people's machines.  Any pain associated is an effective learning tool.

I am _absolutely_ in favor of simplifying things though.  E.g. ideally we'd probably have a "just run `make` prior to PR" instruction, which runs a specific (heavyweight but thorough) Gradle target, and people interested in optimizing that can simply read and find out what the more narrow and faster targets are.

Groxx (changes requested):
Covered in chat internally in a bit more detail, but as a tl;dr: commit hooks are WAY more pain than they're worth, :-1:.  Every single use I've seen has eventually removed them with enthusiasm due to either significant conflicts with reasonable workflows, or absolutely massive numbers of people simply not using them.

If we're missing CI checks: yea, we should _absolutely_ plug those.
If we're too _difficult_ to run CI checks locally, for earlier warning: yea, we should absolutely improve that.
If we're encountering people not reading instructions, and getting caught by CI, and annoyed at the delay: working as intended.  Hooks require reading instructions too, so they won't help with this.  Reading instructions _must_ be assumed, it's better that we can't run arbitrary code on people's machines.  Any pain associated is an effective learning tool.

I am _absolutely_ in favor of simplifying things though.  E.g. personally I find Gradle fairly obtuse, so a trivial Makefile with a few gradle-running targets (this one seems great!) seems reasonable and probably worth doing.

shijiesheng:
already implemented in #783

PR #785, opened 2023-04-20
added MigrationInterceptor class and versioning

abhishekj720:


src/main/java/com/uber/cadence/migration/MigrationInterceptor.java
```diff
@@ -0,0 +1,123 @@
+/*
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
```
Groxx, line 4: hmm.  we should probably check these license-header-files 🤔

src/main/java/com/uber/cadence/client/WorkflowClient.java
```diff
@@ -131,7 +131,7 @@ static WorkflowClient newInstance(IWorkflowService service, WorkflowClientOption
 
   WorkflowClientOptions getOptions();
 
-  IWorkflowService getService();
+  static IWorkflowService getService();
```
Groxx, line 134: this seems like an "almost definitely not correct" change tbh.  this is an extremely large change - is it actually intended?

PR #788, opened 2023-04-27
fix error message for worker factory

duoertai:


shijiesheng (approved):
Thank you for the fix

PR #790, opened 2023-05-02
[TestEnv] populate cronschedule in history start event

shijiesheng:
**Issue**

TestEnv doesn't populate cronschedule in history for cron-workflows. This is blocking testing interceptors. 

**Changes**

* populate cronschedule on the start event
* correct isCron in list workflows

### DIFF
--- a/src/main/java/com/uber/cadence/internal/testservice/StateMachines.java
+++ b/src/main/java/com/uber/cadence/internal/testservice/StateMachines.java
@@ -549,6 +549,7 @@ class StateMachines {
     if (data.continuedExecutionRunId.isPresent()) {
       a.setContinuedExecutionRunId(data.continuedExecutionRunId.get());
     }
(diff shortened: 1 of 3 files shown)


Groxx (approved):
makes sense and seems like an improvement, but two questions:

1. any idea how to prevent this gap in the future?  maybe with some reflection, or even just comments?
2. is there some way to know that this change plugs all existing gaps?  I can fire it up in an IDE and check too, just curious.

shijiesheng:
> makes sense and seems like an improvement, but two questions:
> 
> 1. any idea how to prevent this gap in the future?  maybe with some reflection, or even just comments?
> 2. is there some way to know that this change plugs all existing gaps?  I can fire it up in an IDE and check too, just curious.

1. I added a test to ensure cron schedule is uploaded to starteventattributes. 
2. I think so as state chart is the only place to construct an event.

PR #797, opened 2023-05-08
Expose started event and data converter

abhishekj720:
exposing startedEventAttributes and DataConverter through DecisionContext in WorkflowInfo for use in ContinueAsNew

Added DataConverter get function to change the data to buyte[] format


src/main/java/com/uber/cadence/internal/replay/DecisionContext.java
```diff
@@ -79,6 +80,8 @@ public interface DecisionContext extends ReplayAware {
    */
   SearchAttributes getSearchAttributes();
 
+  DataConverter getDataConverter();
```
shijiesheng, line 83: add comment.

shijiesheng (approved):

shijiesheng (approved):

PR #807, opened 2023-05-17
Added MigrationiWorkflowService and startWorkflow implementation

abhishekj720:
-- Created DummyIWorkflow service implementation of IWorkflowService interface
-- extended to MigrationIWorkflowService
-- Added implementation of startWorkflow based on service in new/old

### DIFF
--- a/src/main/java/com/uber/cadence/migration/MigrationIWorkflowService.java
+++ b/src/main/java/com/uber/cadence/migration/MigrationIWorkflowService.java
@@ -20,6 +20,7 @@ package com.uber.cadence.migration;
 import com.uber.cadence.*;
 import com.uber.cadence.serviceclient.DummyIWorkflowService;
 import com.uber.cadence.serviceclient.IWorkflowService;
+import java.util.Arrays;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import org.apache.thrift.TException;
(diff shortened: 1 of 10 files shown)


src/main/java/com/uber/cadence/migration/MigrationIWorkflowService.java
```diff
@@ -0,0 +1,99 @@
...
+
+    try {
+      serviceNewThread.join();
+      serviceOldThread.join();
+    } catch (InterruptedException e) {
+      e.printStackTrace();
+    }
+
+    // exist in both domains or exist in only new - start in new
+    if (serviceNew != null && serviceOld != null) {
```
shijiesheng, line 93: This is incorrect logic. You are asserting services not the responses.

src/main/java/com/uber/cadence/migration/MigrationIWorkflowService.java
```diff
@@ -0,0 +1,99 @@
...
+
+    Thread serviceNewThread =
+        new Thread(
+            () -> {
+              try {
+                serviceNew.DescribeWorkflowExecution(
+                    new DescribeWorkflowExecutionRequest()
+                        .setDomain(domainNew)
+                        .setExecution(new WorkflowExecution().setWorkflowId(workflowID)));
+              } catch (EntityNotExistsError e) {
```
shijiesheng, line 61: The logic here should suppress the exception

src/main/java/com/uber/cadence/migration/MigrationIWorkflowService.java
```diff
@@ -0,0 +1,99 @@
...
+    return serviceOld.StartWorkflowExecution(startRequest);
+  }
+
+  private boolean shouldStartInNew(String workflowID) {
+
+    Thread serviceNewThread =
+        new Thread(
+            () -> {
+              try {
+                serviceNew.DescribeWorkflowExecution(
```
shijiesheng, line 57: you'll need to assign the response to the executionInNew

PR #820, opened 2023-06-28
Add isolation group header to service client

Shaddoll:
**What?**
Add a new header to thrift and grpc client to specify the isolation group of the client.

### DIFF
--- a/src/main/java/com/uber/cadence/internal/compatibility/proto/serviceclient/GrpcServiceStubs.java
+++ b/src/main/java/com/uber/cadence/internal/compatibility/proto/serviceclient/GrpcServiceStubs.java
@@ -15,6 +15,7 @@
  */
 package com.uber.cadence.internal.compatibility.proto.serviceclient;
 
+import com.google.common.base.Strings;
 import com.uber.cadence.api.v1.DomainAPIGrpc;
 import com.uber.cadence.api.v1.MetaAPIGrpc;
 import com.uber.cadence.api.v1.MetaAPIGrpc.MetaAPIBlockingStub;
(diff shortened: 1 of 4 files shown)


src/main/java/com/uber/cadence/internal/compatibility/proto/serviceclient/GrpcServiceStubs.java
```diff
@@ -61,6 +62,8 @@ final class GrpcServiceStubs implements IGrpcServiceStubs {
       Metadata.Key.of("cadence-client-feature-version", Metadata.ASCII_STRING_MARSHALLER);
   private static final Metadata.Key<String> CLIENT_IMPL_HEADER_KEY =
       Metadata.Key.of("cadence-client-name", Metadata.ASCII_STRING_MARSHALLER);
+  private static final Metadata.Key<String> ISOLATION_GROUP_HEADER_KEY =
+      Metadata.Key.of("cadence-client-isolation-group", Metadata.ASCII_STRING_MARSHALLER);
```
davidporter-id-au, line 66: I suppose i should make the go one match this

davidporter-id-au (approved):
Looks good, but can we ensure we do an e2e test before we merge this

src/main/java/com/uber/cadence/internal/compatibility/proto/serviceclient/GrpcServiceStubs.java
Shaddoll: yeah

src/main/java/com/uber/cadence/serviceclient/ClientOptions.java
```diff
@@ -79,6 +79,8 @@ public class ClientOptions {
   private final IAuthorizationProvider authProvider;
   /** Optional Feature flags to turn on/off some Cadence features */
   private final FeatureFlags featureFlags;
+  /** Optional isolaationGroup of the service */
```
shijiesheng, line 82: nit: typo

src/main/java/com/uber/cadence/internal/compatibility/proto/serviceclient/GrpcServiceStubs.java
shijiesheng: in go client, we didn't make the oss change so it needs two changes. (oss + internal fx package)

shijiesheng (approved):

PR #831, opened 2023-07-25
Fix MethodRetry annotation handling for unlimited maximum attempts

lightoze:


src/main/java/com/uber/cadence/common/RetryOptions.java
```diff
@@ -64,8 +64,11 @@ public static RetryOptions merge(MethodRetry r, RetryOptions o) {
     } else {
       builder.setBackoffCoefficient(DEFAULT_BACKOFF_COEFFICIENT);
     }
+    int maximumAttempts = merge(r.maximumAttempts(), o.getMaximumAttempts(), int.class);
+    if (maximumAttempts != 0) {
```
shijiesheng, line 68: no sure if I follow that. How is zero different from null in the processing? Could you provide the code link?

votez: @shijiesheng I think the point is setMaximumAttempts method throws exception if the value is 0.
The value 0 is default and a valid value if "expiration" is set.
Unfortunately I cannot build the project locally because of thrift 0.9.3 not available on linux currently but test like that should show it:
```java
public class RetryOptionsTest {
  @Test
  public void testUnlimitedRetryWithExpiration() throws NoSuchMethodException {
    RetryOptions topLevelOption =
        new RetryOptions.Builder().setExpiration(Duration.ofSeconds(5)).build();
    MethodRetry methodRetry =
        UnlimitedRetry.class.getMethod("retry").getAnnotation(MethodRetry.class);
    RetryOptions retryOptions = RetryOptions.merge(methodRetry, topLevelOption);
    Assert.assertEquals(
        "Top level options must take precedence", 5, retryOptions.getExpiration().getSeconds());
  }

  private interface UnlimitedRetry {
    @ActivityMethod
    @MethodRetry(expirationSeconds = 100)
    void retry();
  }
}
```

zawadzkidiana (commented):
see below.

zawadzkidiana (commented):
see below

zawadzkidiana (commented):
see below

zawadzkidiana:
Thanks for this, @lightoze, and sorry it's taken this long to get a look.

I checked this against current `master`, and the bug is still there. `RetryOptions.merge(MethodRetry, RetryOptions)` unconditionally calls `builder.setMaximumAttempts(merge(r.maximumAttempts(), o.getMaximumAttempts(), int.class))`. When neither the `@MethodRetry` annotation nor an explicit `RetryOptions` sets `maximumAttempts` — the documented way to ask for "unlimited attempts, bounded only by `expirationSeconds`" — `merge` returns `0`, and `Builder.setMaximumAttempts` throws `IllegalArgumentException("Invalid maximumAttempts: 0")` because it requires `>= 1`. So today, any workflow/activity relying on expiration-only retry via the annotation form fails at merge time, even though the exact same configuration built directly through `RetryOptions.Builder` (skipping `setMaximumAttempts` entirely) works fine — `validate()` itself is perfectly happy with `maximumAttempts == 0`. Your fix — only calling `setMaximumAttempts` when the merged value is non-zero — closes that gap, and it's consistent with how the non-annotation merge path (`RetryOptions.merge(RetryOptions)`, a few lines above) already avoids this by going through the Builder's no-validation constructor path.

The fix itself is correct and minimal. One ask before merge: a regression test in `RetryOptionsTest` that exercises exactly this case — a `@MethodRetry` with `expirationSeconds` set and no `maximumAttempts`, merged against a `null`/default `RetryOptions` — so this doesn't silently regress again. Right now there's nothing pinning this behavior down.

The PR is showing as blocked from merging — worth checking whether that's just missing required reviews/approvals or an actual conflict with `master`; from what I can see the target file hasn't otherwise changed since this was opened, so it should still apply cleanly. Thanks for tracking this down and for your patience — this is a good, focused fix.

---
_Generated by [Claude Code](https://claude.ai/code)_

zawadzkidiana (commented):
see below

zawadzkidiana (changes requested):
Confirmed this bug is real and still present on master. RetryOptions.merge calls builder.setMaximumAttempts(0) when neither side sets maximumAttempts, and Builder.setMaximumAttempts throws on any value under 1. So @MethodRetry with only expirationSeconds set currently crashes at merge time instead of working as documented.

The fix is correct and matches the existing pattern used for backoffCoefficient two lines up. Two things before merge:

1. Add a test. WorkflowOptionsTest already exercises RetryOptions.merge() via @MethodRetry, a case asserting this no longer throws would lock it in.
2. The instance method RetryOptions.merge(RetryOptions) right below has the identical bug (unconditional setMaximumAttempts again). Worth fixing in the same PR since it's the same root cause.

Also: no checks have reported on this branch. GitHub shows it as blocked rather than conflicted, this looks like CI hasn't been triggered at all, possibly pending approval to run Actions for an external contributor's fork rather than a code problem. Worth confirming the test suite actually runs and passes before merge.

PR #837, opened 2023-08-02
Added initialization of contextPropagators from passed options

uberzydins:
Previously `contextPropagators` field was not initialised from passed `WorkflowClientOptions` instance. This pull request presents a fix for that

### DIFF
--- a/src/main/java/com/uber/cadence/client/WorkflowClientOptions.java
+++ b/src/main/java/com/uber/cadence/client/WorkflowClientOptions.java
@@ -59,7 +59,7 @@ public final class WorkflowClientOptions {
     private DataConverter dataConverter = JsonDataConverter.getInstance();
     private WorkflowClientInterceptor[] interceptors = EMPTY_INTERCEPTOR_ARRAY;
     private Scope metricsScope = NoopScope.getInstance();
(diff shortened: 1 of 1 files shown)


src/main/java/com/uber/cadence/client/WorkflowClientOptions.java
```diff
@@ -72,6 +72,7 @@ private Builder(WorkflowClientOptions options) {
       metricsScope = options.getMetricsScope();
       identity = options.getIdentity();
       queryRejectCondition = options.getQueryRejectCondition();
+      contextPropagators = options.getContextPropagators();
```
Groxx, line 75:
```suggestion
      contextPropagators = options.getContextPropagators();
      queryRejectCondition = options.getQueryRejectCondition();
```
extremely minor but might as well keep things in source order

Groxx (approved):
Pretty straightforward.  Thanks!
Got one auto-apply-able thing if ya want, but it's fine either way imo.

src/main/java/com/uber/cadence/client/WorkflowClientOptions.java
uberzydins: Thanks, done

PR #851, opened 2023-08-31
fixing flaky test by increasing the timeout

abhishekj720:
Test: whenHistoryIsPartialCachedEntryIsReturned
- set timeout to 8000


shijiesheng:
it's no longer an issue now

PR #854, opened 2023-09-27
Fix potential concurrent modification in LoggerTest

votez:


src/test/java/com/uber/cadence/workflow/LoggerTest.java
```diff
@@ -118,7 +120,8 @@ public void testWorkflowLogger() {
 
   private int matchingLines(String message) {
     int i = 0;
-    for (ILoggingEvent event : listAppender.list) {
+    List<ILoggingEvent> loggingEvents = new ArrayList<>(listAppender.list);
+    for (ILoggingEvent event : loggingEvents) {
```
Groxx, line 124: seems unrelated, the list is still being iterated over to make the copy -> that can still race with something modifying the list.

votez: this technique is called []defensive copying](http://www.javapractices.com/topic/TopicAction.do?Id=15). _ArrayList_ has a _copying constructor_ so when the original list (from appender) get modified it does not affect the copy so the ConcurrentModificationException is avoided.

Groxx: I am quite bad at keeping up with github notifications, so apologies about the (gigantic) delay on this :\  it was not intended, I didn't notice the response and then it was gone forever.

Defensive copying does not apply in this kind of concurrent modification - what's happening here is that _other threads_ are modifying the list while it's being iterated - that iteration is still happening with this copy, it's just happening inside the ArrayList constructor, so the race is still here.  It has just moved one line earlier.

Defensive copying only solves the case where the contents of the loop modify the collection being looped over.  For that scenario it's useful / a real fix, but that's not occurring here.

#885 has a correct fix: synchronize access to the list itself.  
Ideally we wouldn't have background threads pushing data in here after the test has "ended", but that's unfortunately likely to take a lot more work to solve :\

Groxx:
Gonna close since #885 solves this, so it's no longer needed.

Apologies on the delay!  And thanks for the PR regardless, more are welcome and I'll try to check my notifications more regularly (and we've got some more teammates, hopefully that'll help keep a better eye on the various repos going forward).

votez:
it does solves the problem because the ArrayList constructor is thread-safe of ArrayList source. The original ListAppender is backed by ArrayList and the unit test copies it in ArrayList without involving an iterator, thread-safe again.

But I understand the implementation of logback's ListAppender can change without notice and internals of ArrayList constructor is not obvious for everyone so synchronized could pass review in one day and this only causes doubts and misentrpretation...

PR #858, opened 2023-09-29
use the "test" scope for the io.grpc:grpc-testing dependency

votez:
reduces runtime jar and removes unused classes from the distribution.
I.e. JUnit is not needed in prod


natemort (approved):
LGTM, rebased.

Issue #856, opened 2023-09-28
Client doesn't set sticky kind for sticky tasklist


mantas-sidlauskas:
### Java client is not setting tasklist `kind` property during poll requests

This is causing lots of stikcy tasklist uuids to appear in metrics. This is taking all available memory to keep metrics when using prometheus client. 

```
cadence_latency_per_tl_bucket{cadence_service="cadence_matching",caller="cadence_frontend",domain="samples_domain",operation="CancelOutstandingPoll",tasklist="sticky_Mantass_MacBook_Air_local_4dff8ca1_f2e2_482e_a9a3_8e712f38c8c5",transport="grpc",le="5e-09"} 0
```

Cadence server inspects tasklist kind and converts sticky tasklist name to a special `__sticky__` name when request sends `TaskListKindSticky` property.

Go client, on the other hand, sets  correctly https://github.com/uber-go/cadence-client/blob/bdc3c39797e572f6e2b19ea1b3f3b4e4dff4e786/internal/internal_task_pollers.go#L742

mantas-sidlauskas:
Fixed in https://github.com/uber/cadence-java-client/pull/859

PR #859, opened 2023-09-29
Populate tasklistkind in poll request

shijiesheng:
See https://github.com/uber/cadence-java-client/issues/856

### DIFF
--- a/src/main/java/com/uber/cadence/internal/worker/TaskListKind.java
+++ b/src/main/java/com/uber/cadence/internal/worker/TaskListKind.java
@@ -1,0 +1,33 @@
+/*
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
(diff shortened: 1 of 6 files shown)


src/main/java/com/uber/cadence/internal/worker/TaskListKind.java
```diff
@@ -0,0 +1,33 @@
...
+  TASK_LIST_KIND_STICKY(1);
+
+  private final int value;
+
+  TaskListKind(int value) {
+    this.value = value;
+  }
+
+  public com.uber.cadence.TaskListKind toThrift() {
+    return com.uber.cadence.TaskListKind.findByValue(this.value);
```
davidporter-id-au, line 31: I'm sure this is very borderline value for now, but in case some enums get out of sync or something, unit-test coverage?

shijiesheng: right. added unit test

davidporter-id-au (approved):

PR #867, opened 2024-02-28
modified documentation to not use DataConverter incorrectly

Krashcan:
**What changed?**
Documentation for Worker constructor

**Why?**
It incorrectly mentioned that things like DataConverter are being passed via options which is no longer the case since release v3.0.0.

**How did you test it?**
No actual code change done

**Potential risks**
Nothing, worst case my documentation itself is incorrect.

**Release notes**
No

**Documentation Changes**
No

### DIFF
--- a/src/main/java/com/uber/cadence/worker/Worker.java
+++ b/src/main/java/com/uber/cadence/worker/Worker.java
@@ -23,7 +23,7 @@ import com.google.common.base.Preconditions;
 import com.uber.cadence.client.WorkflowClient;
 import com.uber.cadence.common.WorkflowExecutionHistory;
 import com.uber.cadence.context.ContextPropagator;
-import com.uber.cadence.converter.DataConverter;
+import com.uber.cadence.internal.worker.PollerOptions;
 import com.uber.cadence.internal.common.InternalUtils;
 import com.uber.cadence.internal.metrics.MetricsTag;
 import com.uber.cadence.internal.replay.DeciderCache;
@@ -61,7 +61,7 @@ public final class Worker implements Suspendable {
    * @param client client to the Cadence Service endpoint.
    * @param taskList task list name worker uses to poll. It uses this name for both decision and
    *     activity task list polls.
-   * @param options Options (like {@link DataConverter} override) for configuring worker.
+   * @param options Options (like {@link PollerOptions} override) for configuring worker.
    */
   Worker(
       WorkflowClient client,


zawadzkidiana (commented):
Thanks for the doc cleanup — agreed that Javadoc referencing a mechanism that no longer applies is worse than no comment at all, since it actively misleads whoever goes looking at it next.

Looking at the current `Worker.java` on `master`, though, this specific constructor's Javadoc has since been rewritten as part of a larger change — the constructor signature itself now takes `factoryOptions`, `cache`, `enableStickyExecution`, `threadPoolExecutor`, `contextPropagators`, etc., and the `@param options` line no longer mentions `DataConverter` at all. So this particular one-line fix looks like it's already been superseded and won't apply cleanly against current `master` anymore.

If there's still stale `DataConverter`-via-worker-options wording elsewhere in the docs/Javadoc that you ran into, it'd be great to get an updated PR against current `master` for that — happy to take a look. Otherwise this one is probably safe to close as no-longer-applicable. Thanks for taking the time to send it in either way!

---
_Generated by [Claude Code](https://claude.ai/code)_

zawadzkidiana:
Thanks for catching this, @Krashcan, and sorry it sat here unreviewed for so long.

Your report was accurate. The `Worker` constructor's Javadoc said `@param options Options (like {@link DataConverter} override) for configuring worker`, but `DataConverter` hasn't been settable through `WorkerOptions` since v3.0.0 — the constructor pulls it from the client instead:

```java
SingleWorkerOptions.newBuilder()
    .setDataConverter(client.getOptions().getDataConverter())
```

So the doc was pointing people at a knob that isn't on that dial, which is a genuinely misleading kind of stale comment — worse than no comment.

**Current state on `master`**

The line has since been fixed, though by a different route. The constructor Javadoc now reads:

```java
 * @param options Options for configuring worker.
```

with the `{@link}` dropped entirely and the `com.uber.cadence.converter.DataConverter` import removed. The surrounding `@param` block was also filled out at the same time (`factoryOptions`, `cache`, `enableStickyExecution`, `threadPoolExecutor`, `contextPropagators` all have entries now). So the incorrect reference you flagged is gone.

For what it's worth, I think plain wording was the better landing spot than the substitution proposed here, for a reason worth noting: `PollerOptions` lives in `com.uber.cadence.internal.worker`. Linking it from a public-API Javadoc both adds an import of an internal type and generates a doc link into a package users aren't meant to depend on. Not something you could have been expected to weigh from outside the codebase — and it doesn't diminish the report, which is the part that actually mattered.

The branch also now conflicts with `master` (the file has moved on considerably since February 2024), so there's nothing left to rebase toward.

I'd suggest closing this as resolved. Thank you for reading the docs closely enough to notice, and for taking the trouble to open a PR rather than just working around it — small correctness fixes to public-API docs are genuinely valuable, and I'm sorry this particular one didn't get the timely response it deserved. Please don't let the delay put you off filing the next one.

---
_Generated by [Claude Code](https://claude.ai/code)_

zawadzkidiana (commented):
The specific Javadoc this PR fixes has already been rewritten on master as part of a larger Worker.java change, the @param options line no longer mentions DataConverter at all. This won't apply cleanly anymore.

Recommend closing as no longer applicable. If there's other stale DataConverter-via-worker-options wording still in the docs, a fresh PR against current master would be useful.

Separately: this branch has merge conflicts, which is why no checks have run on it. Not worth rebasing given the doc line is already gone upstream.

PR #878, opened 2024-04-16
Migrate Buildkite CI queues from AWS to GKE

mstifflin:
**What changed?**
- Update Buildkite pipeline yaml to work with the newly provisioned queues in Google Kubernetes Engine
- Use [agent-stack-k8s v0.8.0 helm chart](https://github.com/buildkite/agent-stack-k8s/tree/v0.8.0?tab=readme-ov-file#cloning-repos-via-ssh) which has its own expected pipeline yaml syntax in order to successfully onboard. 
- Install buildkite-agent in Dockerfile for use in the code coverage step
  - The mount that previously worked in AWS's VM based infra doesn't work in Kubernete's container based set up. Install the cli directly for simplicity. 

**Why?**
- There is a company mandate to evacuate Uber's AWS CI footprint. The go forward decision is to use Google Cloud. Buildkite Enterprise recommends GKE (as opposed to GCP) as the way to have a queue with autoscaling compute. 

**How did you test it?**
- https://buildkite.com/uberopensource/cadence-java-client-gcp/builds/19

**Potential risks**
- CI builds will be broken or flaky
- Can be mitigated by a git revert

**Release notes**
n/a

**Documentation Changes**
n/a

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,76 +1,146 @@
+container:
+  kubernetes: &kubernetes
+    gitEnvFrom:
+      - secretRef:
+          name: oss-github-ssh-credentials
+    sidecars:
+    - image: us-west1-docker.pkg.dev/ci-compute/buildkite-images/buildkite-dind:v1
+      volumeMounts:
+        - mountPath: /var/run/
+          name: docker-sock
+      securityContext:
+        privileged: true
+        allowPrivilegeEscalation: true
+    mirrorVolumeMounts: true # CRITICAL: this must be at the same indentation level as sidecars
+    podSpec: &podSpec
+      containers:
+        - &commandContainer
+          image: us-west1-docker.pkg.dev/ci-compute/buildkite-images/buildkite-command-container:v2
+          command:
+          - |-
+            echo "Command step was not overridden."
+            exit 1
+          volumeMounts:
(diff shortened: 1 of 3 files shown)


.buildkite/pipeline.yml
```diff
@@ -1,76 +1,146 @@
...
           run: unit-test-test-service
           config: docker/buildkite/docker-compose.yaml
 
   - label: ":java: Unit test with docker services sticky on"
-    agents:
-      queue: "workers"
-      docker: "*"
-    command: "./gradlew --no-daemon test"
-    timeout_in_minutes: 15
+    timeout_in_minutes: 30
```
taylanisikdemir, line 88: do we expect CI jobs to take longer on GKE?

taylanisikdemir (approved):

davidporter-id-au (approved):

.buildkite/pipeline.yml
mstifflin: It does seem a bit flakier when leaving the timeout as is, so I bumped it. I don't have a concrete root cause, but the current theory is moving from VM based infra on AWS where you have dedicated bandwidth to a k8s cluster with potentially many containers competing for bandwidth could be part of the issue. CPU / Memory resources were configured to closely match what was available on AWS.

PR #880, opened 2024-04-18
Changed default tracer from NoopTracer to GlobalTracer

shijiesheng:
**What changed?**

* Default tracer will use GlobalTracer instead.
* Removed one constructor of `SyncDecisionContext` to set tracer more explicitly so it's less error-prone. 

**Why?**

Simplify onboarding time and match go client behavior. 

**How did you test it?**

Integration test for default behavior

**Potential risks**

**Release notes**

**Documentation Changes**


shijiesheng:
after discussion, it's better to provide tracer in an explicit way

PR #884, opened 2024-04-25
Add initial support for StartWorkflowExecutionAsync

natemort:
**What changed?**
Add WorkflowStub#enqueueStart which uses the new StartWorkflowExecutionAsync to submit a workflow to be started at some point in the future.

Currently nearly all testing of the Cadence Java client is implemented as integration tests in WorkflowTests, running against either a locally running Cadence service or using the test environment. While this approach would be technically possible to configure (adding zookeeper and kafka to the docker-compose files), this would require reimplementing the integration tests already present on the Cadence server for all new features added. This approach doesn't scale well for adding new functionality to each client.

Instead, we add tests for this functionality via targeted unit tests at the different layers of the client. Add WorkflowClientInternalTest to use an in-memory fake server with stubbed responses to assert that we make exactly the correct request. This tests the entire Tchannel networking stack of the client end to end. Add WorkflowStubImplTest to test the specific semantics and state of the WorkflowStubImpl. State is particularly important with WorkflowStubs as they're intended to represent a single execution of a Workflow, so they normally capture the Workflow runId when starting a workflow. For StartWorkflowExecutionAsync we capture only the workflowId so that a stub can be used to enqueueStart(), then later qury or signal that workflow.

Additionally add tests for the thrift2proto mappers, including the MapperTestUtil assertions to ensure that all fields are present. This ensures that any update to the IDL files requires a corresponding change to either the test or mapper to account for the new field.

**Why?**
- Adding support for new Cadence features to the Java client

**How did you test it?**
- Unit tests
- Running WorkflowTest against locally running cadence instead with async workflows enabled

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/src/main/java/com/uber/cadence/client/WorkflowStub.java
+++ b/src/main/java/com/uber/cadence/client/WorkflowStub.java
@@ -76,11 +76,12 @@ public interface WorkflowStub {
   CompletableFuture<WorkflowExecution> startAsyncWithTimeout(
       long timeout, TimeUnit unit, Object... args);
 
-  void enqueueStart(Object... args);
+  WorkflowExecution enqueueStart(Object... args);
 
-  CompletableFuture<Void> enqueueStartAsync(Object... args);
+  CompletableFuture<WorkflowExecution> enqueueStartAsync(Object... args);
 
-  CompletableFuture<Void> enqueueStartAsyncWithTimeout(long timeout, TimeUnit unit, Object... args);
+  CompletableFuture<WorkflowExecution> enqueueStartAsyncWithTimeout(
+      long timeout, TimeUnit unit, Object... args);
 
   WorkflowExecution signalWithStart(String signalName, Object[] signalArgs, Object[] startArgs);
 
--- a/src/main/java/com/uber/cadence/client/WorkflowStub.java
+++ b/src/main/java/com/uber/cadence/client/WorkflowStub.java
@@ -76,6 +76,12 @@ public interface WorkflowStub {
   CompletableFuture<WorkflowExecution> startAsyncWithTimeout(
       long timeout, TimeUnit unit, Object... args);
 
+  void enqueueStart(Object... args);
+
+  CompletableFuture<Void> enqueueStartAsync(Object... args);
+
+  CompletableFuture<Void> enqueueStartAsyncWithTimeout(long timeout, TimeUnit unit, Object... args);
+
   WorkflowExecution signalWithStart(String signalName, Object[] signalArgs, Object[] startArgs);
 
   Optional<String> getWorkflowType();
--- a/src/main/java/com/uber/cadence/internal/compatibility/proto/serviceclient/GrpcServiceStubs.java
+++ b/src/main/java/com/uber/cadence/internal/compatibility/proto/serviceclient/GrpcServiceStubs.java
@@ -193,54 +193,57 @@ final class GrpcServiceStubs implements IGrpcServiceStubs {
             span.finish();
           }
 
+          @SuppressWarnings("unchecked")
           @Override
           public void sendMessage(ReqT message) {
             if (Objects.equals(method.getBareMethodName(), "StartWorkflowExecution")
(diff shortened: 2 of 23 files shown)


src/main/java/com/uber/cadence/serviceclient/WorkflowServiceTChannel.java
```diff
@@ -663,6 +649,66 @@ private StartWorkflowExecutionResponse startWorkflowExecution(
...
+      if (result.isSetDomainNotActiveError()) {
+        throw result.getDomainNotActiveError();
+      }
+      if (result.isSetLimitExceededError()) {
+        throw result.getLimitExceededError();
+      }
+      if (result.isSetEntityNotExistError()) {
+        throw result.getEntityNotExistError();
+      }
+      throw new TException("StartWorkflowExecution failed with unknown error:" + result);
```
shijiesheng, line 695: missing check on clientversion not supported error. This bug existed in the original sync apis

src/main/java/com/uber/cadence/client/WorkflowStub.java
```diff
@@ -76,6 +76,12 @@ CompletableFuture<Void> signalAsyncWithTimeout(
   CompletableFuture<WorkflowExecution> startAsyncWithTimeout(
       long timeout, TimeUnit unit, Object... args);
 
+  void enqueueStart(Object... args);
```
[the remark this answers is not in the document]
natemort: Currently we only have documentation on the main WorkflowClient [methods](https://github.com/uber/cadence-java-client/blob/master/src/main/java/com/uber/cadence/client/WorkflowClient.java#L359) and on the class description for WorkflowClient. I was thinking it would make sense to add it there to be more in line with our current practices.

I'm planning on a followup diff to add in all the overloads to WorkflowClient to try and reduce the scope of this. Adding support to WorkflowStub is just the "minimum" needed to get this working end to end.

src/main/java/com/uber/cadence/serviceclient/WorkflowServiceTChannel.java
natemort: Fixed. I thought this might be a new error that none of the endpoints accounted for but it seems that StartWorkflowExecution is actually the only one missing it. Nice catch.

src/test/java/com/uber/cadence/internal/sync/WorkflowClientInternalTest.java
```diff
@@ -0,0 +1,95 @@
...
+import java.nio.charset.StandardCharsets;
+import java.time.Duration;
+import java.util.concurrent.CompletableFuture;
+import org.junit.Before;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+public class WorkflowClientInternalTest {
+
+  @ClassRule public static FakeWorkflowServiceRule fakeService = new FakeWorkflowServiceRule();
```
shijiesheng, line 23: use mockito which is easier?

src/main/java/com/uber/cadence/internal/compatibility/proto/serviceclient/GrpcServiceStubs.java
```diff
@@ -193,55 +193,58 @@ public void start(Listener<RespT> responseListener, Metadata headers) {
...
-                    headerBuilder.putFields(
-                        k, Payload.newBuilder().setData(ByteString.copyFrom(v)).build());
-                  });
+              Header newHeader = addTracingHeaders(request.getHeader());
+
+              // cast should not throw error as we are using the builder
+              message = (ReqT) request.toBuilder().setHeader(newHeader).build();
+            } else if (Objects.equals(method.getBareMethodName(), "StartWorkflowExecutionAsync")
+                && message instanceof StartWorkflowExecutionAsyncRequest) {
+              StartWorkflowExecutionAsyncRequest request =
```
shijiesheng, line 208: could you add test to assert trace is set on `StartWorkflowExecutionAsync`

shijiesheng (approved):

PR #885, opened 2024-04-25
Fix flakiness in LoggerTest

natemort:
**What changed?**
If any other threads logs a message while we iterate over the events it can throw a ConcurrentModificationException.

Create our own implementation of ListAppender that uses the synchronized keyword to protect reading/writing to the backing list with a mutex. Create a copy of the list while holding to the lock to support iterating over it.

**Why?**
- Fix flaky test

**How did you test it?**
- Running the test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/src/test/java/com/uber/cadence/workflow/LoggerTest.java
+++ b/src/test/java/com/uber/cadence/workflow/LoggerTest.java
@@ -22,7 +22,7 @@ import static org.junit.Assert.assertTrue;
 
 import ch.qos.logback.classic.LoggerContext;
 import ch.qos.logback.classic.spi.ILoggingEvent;
-import ch.qos.logback.core.read.ListAppender;
+import ch.qos.logback.core.AppenderBase;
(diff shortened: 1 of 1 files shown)


Groxx:
Simple and should solve the problem 👍 

I am curious what the actual racing writer was though, if you found it.  I have a vague feeling it's probably reasonable and this is truly the correct fix / a pragmatic tradeoff vs rewriting a ton, I'm just not familiar enough with this codebase to know if that's _actually_ the case or not.

Groxx (approved):

natemort:
Here's the output from a local failure: https://gist.github.com/natemort/b8321dcc3d1abe153f37e6c2e7029f9b

I think it's messages from a background thread from the previous test. I see this spam happen quite a bit with the TestWorkflowEnvironment. I'm hoping to fix it at some point.

Groxx:
yea, that's not surprising at all, sadly.

makes sense / fix is correct, thanks!

PR #890, opened 2024-05-03
Lock TestWorkflowStoreImpl when listing workflows

natemort:
This is the only path to the backing histories map that isn't locked. When running a test containing multiple child workflows in parallel where one child workflow attempts to call listWorkflows I'm able to consistently reproduce a concurrent modification exception.

**What changed?**
- Correct locking for ListWorkflows* calls when using the Test Environment

**Why?**
- Fix bug encountered only in unit tests.

**How did you test it?**
- Unit tests

**Potential risks**
Despite being a change to locking, this change is rather low risk. The lock is reentrant and there are no other locks acquired while holding this lock, so there's no chance of deadlock. In addition, this codepath isn't use internally at all. It exists only to implement ListOpenWorkflowExecutions and ListClosedWorkflowExecutions on the test environment, allowing test code to make these RPCs. This is likely why it hasn't been caught before.

**Release notes**

**Documentation Changes**


taylanisikdemir (approved):

abhishekj720 (approved):
added just a try/catch with lock.  LGTM. thank you for the change.

PR #897, opened 2024-05-15
Refactor ActivityTaskExecutor to use ActivityTask interface

shijiesheng:
**What changed?**

Correctly use ActivityTask interface rather than the implementation directly in the interface.

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


natemort (approved):
LGTM, makes sense. It's interesting that we didn't include the input in ActivityTask before but I don't see any harm in doing so.

taylanisikdemir (approved):

PR #898, opened 2024-05-15
Add unsetCurrentContext to ContextPropagator interface

shijiesheng:
**What changed?**

* added unsetCurrentContext method to ContextPropagator

**Why?**

Existing ContextPropagator interface do not support context deactivation

**How did you test it?**

unit test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/src/test/java/com/uber/cadence/internal/testing/WorkflowTestingTest.java
+++ b/src/test/java/com/uber/cadence/internal/testing/WorkflowTestingTest.java
@@ -17,9 +17,7 @@
 
 package com.uber.cadence.internal.testing;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
 import static org.mockito.Matchers.anyString;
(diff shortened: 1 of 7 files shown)


src/test/java/com/uber/cadence/internal/testing/WorkflowTestingTest.java
```diff
@@ -775,6 +775,11 @@ public Object getCurrentContext() {
     public void setCurrentContext(Object context) {
       MDC.put("test", String.valueOf(context));
     }
+
+    @Override
+    public void unsetCurrentContext() {
+      MDC.remove("test");
```
natemort, line 781: Nit: Maybe add an assertion that this key is removed

natemort (approved):

taylanisikdemir (approved):

src/test/java/com/uber/cadence/internal/testing/WorkflowTestingTest.java
shijiesheng: sure. There is actually no easy way to assert on MDC directly, so I added a counter to ensure state is cleaned.

PR #906, opened 2024-05-24
Fix context propagation bug that would link two parents in some cases

shijiesheng:
**What changed?**

* ignore active span in ExecuteWorkflow and ExecuteActivity.

**Why?**

Because workflowThreads are coroutines, it's possible that two workflows are referencing each other as parents.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/src/main/java/com/uber/cadence/internal/tracing/TracingPropagator.java
+++ b/src/main/java/com/uber/cadence/internal/tracing/TracingPropagator.java
@@ -102,6 +102,7 @@ public class TracingPropagator {
     Span span =
         tracer
             .buildSpan(EXECUTE_LOCAL_ACTIVITY)
+            .ignoreActiveSpan()
             .addReference(References.FOLLOWS_FROM, parent)
             .withTag(TAG_WORKFLOW_ID, params.getWorkflowExecution().getWorkflowId())
             .withTag(TAG_WORKFLOW_RUN_ID, params.getWorkflowExecution().getRunId())
--- a/src/main/java/com/uber/cadence/internal/tracing/TracingPropagator.java
+++ b/src/main/java/com/uber/cadence/internal/tracing/TracingPropagator.java
@@ -54,7 +54,7 @@ public class TracingPropagator {
   }
 
   public Span spanByServiceMethod(String serviceMethod) {
-    return tracer.buildSpan(serviceMethod).asChildOf(tracer.activeSpan()).start();
+    return tracer.buildSpan(serviceMethod).start();
   }
 
   public Span spanForExecuteWorkflow(DecisionContext context) {
--- a/src/main/java/com/uber/cadence/internal/tracing/TracingPropagator.java
+++ b/src/main/java/com/uber/cadence/internal/tracing/TracingPropagator.java
@@ -64,7 +64,8 @@ public class TracingPropagator {
 
     return tracer
         .buildSpan(EXECUTE_WORKFLOW)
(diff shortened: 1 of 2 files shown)


src/main/java/com/uber/cadence/internal/tracing/TracingPropagator.java
```diff
@@ -64,6 +64,8 @@ public Span spanForExecuteWorkflow(DecisionContext context) {
 
     return tracer
         .buildSpan(EXECUTE_WORKFLOW)
+        .ignoreActiveSpan() // ignore active span to start a new trace that ONLY links the start
```
quessliu, line 67: Hmm, are we sure this does what we want it to do? Ignoring the current span will prevent two coroutines running on the same thread from inadvertently picking up each others' context state, but then how/where do we track the context state if not on the threadlocal?

quessliu:
Is there any way we can validate this with a unit test that fails without this change and passes with it? I.e. setup a scenario where we initialize a span on the threadlocal, have multiple workflows running on the same thread, and verify that they don't grab the span.

Groxx:
I _suspect_ this is right - we're ignoring the thread-local span because all three of these are run in worker threads, right?

I'm not sure I'm able to validate this without spending a lot of time reading/learning/testing it out though :\

src/main/java/com/uber/cadence/internal/tracing/TracingPropagator.java
shijiesheng: It's linked explicitly by `asChild`

shijiesheng:
> I _suspect_ this is right - we're ignoring the thread-local span because all three of these are run in worker threads, right? With a possible exception for `spanForExecuteWorkflow` since I was under the impression that we run a new thread per workflow, not a shared worker thread (reused after a workflow ends). But I'm not sure if that's true or not.
> 
> I'm not sure I'm able to validate this without spending a lot of time reading/learning/testing it out though, and I don't think I can spend that right now :\

I tried to start 100 workflows with 2 workflow threads but still cannot catch this issue. But the idea is to link workflow trace explicitly ONLY to the start workflow trace.

shijiesheng:
> Is there any way we can validate this with a unit test that fails without this change and passes with it? I.e. setup a scenario where we initialize a span on the threadlocal, have multiple workflows running on the same thread, and verify that they don't grab the span.

I tried by starting 100 workflows with only 2 workflow threads but still cannot catch this.

src/main/java/com/uber/cadence/internal/tracing/TracingPropagator.java
shijiesheng: It's linked explicitly by `asChild`

src/main/java/com/uber/cadence/internal/tracing/TracingPropagator.java
```diff
@@ -54,7 +54,7 @@ public TracingPropagator(Tracer tracer) {
   }
 
   public Span spanByServiceMethod(String serviceMethod) {
-    return tracer.buildSpan(serviceMethod).asChildOf(tracer.activeSpan()).start();
```
shijiesheng, line 57: This asChildOf is unnecessary.

Groxx:
Approving to unblock since I think this is likely correct, we can just try this out in prod with a full stack if needed.

I would love us to get a jaeger collector test suite set up at some point, for both clients :\  changing this stuff is far too fuzzy.

Groxx (approved):

Issue #647, opened 2021-10-26
Java 17 support - Illegal reflective access in CustomThrowableTypeAdapter and CheckedExceptionWrapper


WellingR:
I tried using the cadence-java-client with java 17. However it appears that the `CustomThrowableTypeAdapter` uses a reflection method which no longer works in java 17.

The code
https://github.com/uber/cadence-java-client/blob/master/src/main/java/com/uber/cadence/converter/CustomThrowableTypeAdapter.java#L83
Makes the cause field accessible in order to set a non-accessible field to null. This is no longer possible in java 17 because of https://openjdk.java.net/jeps/403 .

Similarly, the initializer of [CheckedExceptionWrapper](https://github.com/uber/cadence-java-client/blob/master/src/main/java/com/uber/cadence/internal/common/CheckedExceptionWrapper.java#L35) also uses a reflective method which is no longer allowed.

longquanzheng:
It seems like there are some standard replacements recommended. Would you like to propose some changes to that?

jontro:
Is it possible to use the same solution as in temporal? https://github.com/temporalio/sdk-java/commit/4bb23c3a660de6a2630901d0dc6c049d9742ebe1

Also handling errors seems like a nice fix: https://github.com/temporalio/sdk-java/commit/6653b3d71af118824f5eedcbef6e33be103d9f1e

agarwall-rohan:
Is there any solve or alternative proposed for this issue for using uber cadence with Java-17? I am facing java.lang.NoClassDefFoundError: Could not initialize class com.uber.cadence.internal.common.CheckedExceptionWrapper after upgrading to java 17. This should be related to the issue stated above.

sodre90:
If you include the VM parameter "--add-opens java.base/java.lang=ALL-UNNAMED," it will prevent the exception from being thrown.

PR #914, opened 2024-07-09
[Breaking Change] Remove Reflection from CheckedExceptionWrapper

natemort:
Throwable (which Exception and Error are subclasses of)is the base class of anything that can be thrown/caught, and maintains a chain of causality between the Throwable instances, analogous  to Go's error wrapping. Unlike Go, Java has a hierarchy of Throwable types and a concept of "checked exceptions", which are the types of Throwables that have to be declared to be allowed to be thrown within a method.

The Cadence Java Client forbids Workflow interfaces from declaring any checked exceptions and instead requires them to be wrapped via `Workflow.wrap`, which calls through to `CheckedExceptionWrapper.wrap`. When reporting these exceptions back to the Cadence service the client uses `CheckedExceptionWrapper.unwrap` to remove all CheckedExceptionWrapper instances from the chain .

`CheckedExceptionWrapper.unwrap` currently uses reflection to rewrite the chain of Exceptions to remove all instances of CheckedExceptionWrapper. This is an uncommon approach, and doesn't work well with the encapsulation in Java 17. Rather than trying to work around it, remove the behavior.

RpcRetryer was particularly impacted because both it and CompletableFuture wrap exceptions. If looking at only the top of the chain it would see a CompletionException that wrapped a CheckedExceptionWrapper rather than the CheckedExceptionWrapper. CompletableFuture would then unwrap the exception and expose the CheckedExceptionWrapper to the caller within the ExecutionException. Refactor it to avoid wrapping exceptions at all but still unwrap any exceptions thrown by user code so that it remains the same.

This approach largely matches the approach Temporal took in https://github.com/temporalio/sdk-java/pull/251 . They additionally removed `Workflow.unwrap` as there isn't any reason for user code to be calling it (we always throw unwrapped exceptions, so it's a no-op), but I'd rather avoid that breaking change.

**What changed?**
- `CheckedExceptionWrapper.unwrap` no longer mutates Throwables within the causality chain to remove all instances of CheckedExceptionWrapper. It only unwraps the top exception.

**Why?**
- Better support Java 17. See #647 

**How did you test it?**
- Unit tests

**Potential risks**
- Users that had code which caught an Exception, wrapped it with CheckedExceptionWrapper, and then re-wrapped it will see a different Exception chain which still includes the CheckedExceptionWrapper. It's unlikely that their logic relies on this relationship, as they would have to be relying on deeply nested exceptions. For this change to impact their behavior in the case of an activity throwing an exception, they would need a strong dependency on WorkflowFailureException with cause ActivityFailureException with cause (User thrown Exception A) with cause (User Caught Exception B), where they wrapped B in CheckedExceptionWrapper, then wrapped it again in Exception A, and relied on the CheckedExceptionWrapper being removed.

**Release notes**

**Documentation Changes**


src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -6251,4 +6260,43 @@ public void testEnqueueWorkflow() throws Exception {
...
+              TestActivities.class,
+              new ActivityOptions.Builder()
+                  .setTaskList(Workflow.getWorkflowInfo().getTaskList())
+                  .build());
+      activity.throwIOError();
+    }
+  }
+
+  @Test
+  public void testActivityThrowsError() throws Exception {
```
natemort, line 6287: This was mostly for my own curiosity since we have special handling for this case and currently no test coverage. We intentionally don't report Errors, which I'm not convinced is a good or useful thing.

taylanisikdemir (approved):

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
Groxx: given the severity of what `Error` generally represents, tbh I think "have no special handling, let it tear things apart as it goes" is likely best.

natemort:
Will revisit with broader exception improvements in the future

PR #917, opened 2024-07-30
Addition of logs to scanWorkflows in shadower

abhishekj720:
**What changed?**
Added logs to scanWorflow methods

**Why?**
- Adding it to check what query has been passed and has been broken in the process.
- When running shadower, it says invalid query. So, adding this will tell where the problem is.

**How did you test it?**
Just adding logs. Will be adding the tests in later diffs.

**Potential risks**
N/A, as these are just few logs.


agautam478:
LGTM. Approving because we urgently need these logs to debug the issues with internal shadower.

abhishekj720:
Some build issue.

PR #921, opened 2024-10-17
Remove unused code from cadence.internal package

natemort:
**What changed?**
Remove unused code from internal packages to improve code coverage and reduce maintenance burden. Users shouldn't be accessing internals.

**Why?**
Improving code coverage and reducing maintenance burden

**How did you test it?**
Unit tests

**Potential risks**
User code may be accessing internals within the client and could break. Generally this should be a compile time breakage and we do not support accessing internals.

**Release notes**

**Documentation Changes**

### DIFF
--- a/src/main/java/com/uber/cadence/internal/common/WorkflowExecutionUtils.java
+++ b/src/main/java/com/uber/cadence/internal/common/WorkflowExecutionUtils.java
@@ -29,8 +29,6 @@ import com.google.gson.JsonPrimitive;
 import com.uber.cadence.ActivityType;
 import com.uber.cadence.Decision;
 import com.uber.cadence.DecisionType;
-import com.uber.cadence.DescribeWorkflowExecutionRequest;
-import com.uber.cadence.DescribeWorkflowExecutionResponse;
 import com.uber.cadence.EntityNotExistsError;
 import com.uber.cadence.EventType;
 import com.uber.cadence.GetWorkflowExecutionHistoryRequest;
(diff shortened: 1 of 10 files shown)


3vilhamster (approved):

3vilhamster (commented):
I'm unsure if we want to delete this, but I could not find usages of this in monorepos or in open sourced usages.

src/main/java/com/uber/cadence/internal/common/WorkflowExecutionUtils.java
```diff
@@ -621,26 +437,6 @@ public static GetWorkflowExecutionHistoryResponse getHistoryPage(
     return history;
   }
 
-  /** Returns workflow instance history in a human readable format. */
-  public static String prettyPrintHistory(
```
3vilhamster, line 625: This could be useful; I wonder why have you decided to drop this.

src/main/java/com/uber/cadence/internal/sync/WorkflowInternal.java
```diff
@@ -228,26 +220,6 @@ public static <T> T newContinueAsNewStub(
...
-  /**
-   * Execute activity by name.
-   *
-   * @param name name of the activity
-   * @param resultClass activity return type
-   * @param args list of activity arguments
-   * @param <R> activity return type
-   * @return activity result
-   */
-  public static <R> R executeActivity(
```
3vilhamster, line 240: I wonder why is this removed?

Shaddoll (approved):

src/main/java/com/uber/cadence/internal/common/WorkflowExecutionUtils.java
natemort: I think it's potentially useful but it's long unused and untested. If we need similar functionality we can rewrite it.

src/main/java/com/uber/cadence/internal/sync/WorkflowInternal.java
natemort: I'm not really sure why this exists, it's long unused and untested. I think it's best we just remove it.

PR #923, opened 2024-10-22
Refactor Test Environment and Worker initialization to CadenceTestRule

natemort:
**What changed?**
- Refactor workflow initialization logic from WorkflowTest to CadenceTestRule, which can easily be used within in any test. 
  - The simplest example so far is in CrossDomainWorkflowTest, if you ignore that it's using two of them (and the withTestEnvironmentProvider nonsense).
- When run tests are run with USE_DOCKER_SERVICE=true they will run against a real Cadence instance. This is currently run as part of our CI.
- When debugging workflow executions you can change TestEnvironment.DEBUGGER_TIMEOUTS to significantly increase test, workflow, and activity timeouts.
- Tests can be annotated with @RequiresDockerService or @RequiresTestService to only execute in that environment. This also makes it much easier to find behavior disparity or gaps in our test coverage.

WorkflowTest is additionally parameterized by whether or not to use sticky execution. I think there are obviously cases where that makes sense but I don't think we need to do that in every test going forward.

**Why?**
- Allow new tests to be easily written in separate files from the monolithic WorkflowTest.
- Simplify the initialization logic that tests need to worry about to handle running against both the test service and the real service.

**How did you test it?**
- Ran unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/src/test/java/com/uber/cadence/workflow/WorkflowTest.java
+++ b/src/test/java/com/uber/cadence/workflow/WorkflowTest.java
@@ -17,6 +17,7 @@
 
 package com.uber.cadence.workflow;
 
+import static com.uber.cadence.testUtils.TestEnvironment.DOMAIN;
 import static com.uber.cadence.worker.NonDeterministicWorkflowPolicy.FailWorkflow;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
@@ -36,7 +37,6 @@ import com.uber.cadence.CancellationAlreadyRequestedError;
 import com.uber.cadence.DomainAlreadyExistsError;
 import com.uber.cadence.DomainNotActiveError;
 import com.uber.cadence.EntityNotExistsError;
-import com.uber.cadence.FeatureFlags;
 import com.uber.cadence.GetWorkflowExecutionHistoryResponse;
 import com.uber.cadence.HistoryEvent;
 import com.uber.cadence.Memo;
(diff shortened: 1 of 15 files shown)


src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -2464,15 +2288,6 @@ public void mySignal(String value) {
 
   @Test
   public void testSignal() {
-    // Test getTrace through replay by a local worker.
```
natemort, line 2467: This block is a no-op, it create a new Worker and WorkerFactory but doesn't start it.

src/test/java/com/uber/cadence/workflow/WorkflowTest.java
```diff
@@ -3216,12 +3006,7 @@ public WorkflowStub newUntypedWorkflowStub(
                 })
             .setDomain(DOMAIN)
             .build();
-    WorkflowClient wc;
-    if (useExternalService) {
-      wc = WorkflowClient.newInstance(service, clientOptions);
-    } else {
-      wc = testEnvironment.newWorkflowClient(clientOptions);
-    }
+    WorkflowClient wc = cadenceTestRule.createWorkflowClient(clientOptions);
```
shijiesheng, line 3009: nit: maybe add client options to builder?

natemort: Added, good call. This test still needs to create a custom client to test out its interceptor but that's a very useful thing to expose.

shijiesheng (approved):

PR #928, opened 2024-10-25
Report child workflow started synchronously in Test Implementation

natemort:
By reporting child workflow started asynchronously we can observe a child workflow attempting to complete prior to its state machine registering that it started. This is an invalid state transition and causes test failures.

See https://github.com/uber/cadence-java-client/actions/runs/11508671742/job/32037318976 as the most recent occurrence of this issue.

Temporal addressed this issue in https://github.com/temporalio/sdk-java/pull/1289 due to the same test case failing..

**What changed?**
- Fixed cause of test failures

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/src/main/java/com/uber/cadence/internal/testservice/TestWorkflowMutableStateImpl.java
+++ b/src/main/java/com/uber/cadence/internal/testservice/TestWorkflowMutableStateImpl.java
@@ -1179,17 +1179,13 @@ class TestWorkflowMutableStateImpl implements TestWorkflowMutableState {
               .setWorkflowExecution(getExecutionId().getExecution())
               .setDomain(getExecutionId().getDomain())
               .setWorkflowType(startRequest.getWorkflowType());
-      ForkJoinPool.commonPool()
-          .execute(
-              () -> {
-                try {
-                  parent.get().childWorkflowStarted(a);
-                } catch (EntityNotExistsError | WorkflowExecutionAlreadyCompletedError e) {
-                  // Not a problem. Parent might just close by now.
-                } catch (BadRequestError | InternalServiceError e) {
-                  log.error("Failure reporting child completion", e);
-                }
-              });
+      try {
+        parent.get().childWorkflowStarted(a);
+      } catch (EntityNotExistsError | WorkflowExecutionAlreadyCompletedError e) {
+        // Not a problem. Parent might just close by now.
+      } catch (BadRequestError | InternalServiceError e) {
+        log.error("Failure reporting child completion", e);
+      }
     }
   }
 


src/main/java/com/uber/cadence/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -1179,17 +1179,13 @@ public void startWorkflow(
...
-                try {
-                  parent.get().childWorkflowStarted(a);
-                } catch (EntityNotExistsError | WorkflowExecutionAlreadyCompletedError e) {
-                  // Not a problem. Parent might just close by now.
-                } catch (BadRequestError | InternalServiceError e) {
-                  log.error("Failure reporting child completion", e);
-                }
-              });
+      try {
+        parent.get().childWorkflowStarted(a);
```
Groxx, line 1183: huh.  any idea why it was async in the first place?  since it seems to be creating decision tasks above, it seems like the start-event probably already exists, so this doesn't seem to risk deadlocking or something...

natemort: There are two tiers of locks for the test service:
- There's a lock in TestWorkflowService, which controls access to the Map of MutableState for every workflow.
- Every MutableState has a lock.

Typically we lock the TestWorkflowService, get the MutableState, unlock the TestWorkflowService, and then perform an action on the MutableState. The MutableState then manages its own lock internally.

Any time a MutableState performs an action on another workflow we use the ForkJoinPool to run the operation asynchronously without holding any locks for the original workflow. These calls use the TestWorkflowService, so they repeat the process above. This is particularly important for any of the methods that process Decisions, which are all called while the lock for the workflow is held. If two workflows attempted to cancel each other they could deadlock.

By running this block asynchronously we end up triggering a race between:
- This block running and updating the parent workflow that the child workflow started.
- The next decision task being scheduled , processed, and reported back to the TestService, which could complete the child workflow. This similarly prompts an async operation to report the status update to the parent workflow. We see the test failure any time this process beats the first one.
  * The next decision task gets pushed to a queue before we even release the lock on the MutableState.

This block is unique for two reasons:
- Starting a workflow is the only time we hold both the TestWorkflowService lock and MutableState lock simultaneously, so we're still holding the TestWorkflowService lock.
- We've already released the MutableState lock above.

Running this synchronously somewhat incidentally ends up fixing the race condition because we're still holding the lock on the TestWorkflowService. The next decision task can't retrieve the MutableState for the child workflow until we release the lock, so we've created a strict order of events.

Groxx: Makes sense / fits about what I expected, yea.  Thanks!
And it looks like the `childWorkflowStarted` acquires its own lock, so that sounds like a probably-correct fix.

I suppose a later step for all this would be to make these related test workflows share a single execution thread, since otherwise this seems prone to this kind of race for no benefit :\  though I'm gonna assume that's extremely hard without a complete rewrite.

PR #930, opened 2024-10-28
Fix ActivityCompletionClient cancellation and failure by WorkflowExecution

natemort:
When sending RespondActivityTaskFailedByIDRequest or RespondActivityTaskCanceledByIDRequest we don't include the ActivityID. Correctly include the id and add test coverage.

**What changed?**
- Include the ActivityID when canceling/failing activities by ID.

**Why?**
- Fix longstanding bug.
- Increase test coverage.

**How did you test it?**
- Unit/Integration tests.

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/src/test/java/com/uber/cadence/workflow/ManualActivityCompletionWorkflowTest.java
+++ b/src/test/java/com/uber/cadence/workflow/ManualActivityCompletionWorkflowTest.java
@@ -1,0 +1,249 @@
+/*
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"). You may not
+ *  use this file except in compliance with the License. A copy of the License is
+ *  located at
+ *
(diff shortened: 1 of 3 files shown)


src/test/java/com/uber/cadence/workflow/ManualActivityCompletionWorkflowTest.java
```diff
@@ -0,0 +1,249 @@
+/*
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
davidporter-id-au, line 2: Is this correct?

davidporter-id-au: oh, I didn't see the "modifiications copyright" down below, I guess it is from there. the licence URL looks sus, but I guess it is what it is.

natemort: The cadence java client is a fork of https://github.com/aws/aws-swf-flow-library 😄

davidporter-id-au (approved):
Not super strong with the Java impl, but the test looks reasonable from what I can see

PR #937, opened 2024-11-01
Fix flakiness in ManualActivityCompletionWorkflowTest

natemort:
The operations here are flaky because we're starting an activity and then attempting to cancel/complete/fail it in parallel. We need the second activity to block until the first one has started, and there's no way to orchestrate that within the workflow. The best we can do is make the acitivites block until the ActivityTask is available.

**What changed?**
- Fix flaky test

**Why?**
- To reduce wasted effort rerunning tests.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


Groxx (approved):
Yep, makes a lot of sense (assuming it doesn't deadlock or something).  thanks!

PR #944, opened 2024-11-05
adds coverage to SignalExternalWorkflowParameters struct file's methods

davidporter-id-au:
**What changed?**
Adds a small amount of coverage to some internal data object. Fixes up the toString since it seemed to be broken. 

**Why?**
We're trying to lift the coverage and quality of these repos. Incidentally I'm trying to remember how to write Java. 

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/src/main/java/com/uber/cadence/internal/replay/SignalExternalWorkflowParameters.java
+++ b/src/main/java/com/uber/cadence/internal/replay/SignalExternalWorkflowParameters.java
@@ -94,10 +94,20 @@ public final class SignalExternalWorkflowParameters implements Cloneable {
 
   @Override
   public String toString() {
+
+    String inputStr;
+    if (input == null) {
+      inputStr = "";
+    } else if (input.length > 512) {
+      inputStr = new String(input, 0, 512, StandardCharsets.UTF_8);
+    } else {
+      inputStr = new String(input, StandardCharsets.UTF_8);
+    }
+
     StringBuilder sb = new StringBuilder();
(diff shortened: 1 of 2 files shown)


src/main/java/com/uber/cadence/internal/replay/SignalExternalWorkflowParameters.java
```diff
@@ -94,10 +94,20 @@ public SignalExternalWorkflowParameters withWorkflowId(String workflowId) {
 
   @Override
   public String toString() {
+
```
davidporter-id-au, line 97: this whole thing seems utterly pointless and not good, but fixing it up since god forbid, someone mgiht actually be caring about it's current not good structure. 

the main issue is that before, it broke for short strings, which is probably not what you want for a tostring

src/main/java/com/uber/cadence/internal/replay/SignalExternalWorkflowParameters.java
```diff
@@ -108,6 +118,7 @@ public SignalExternalWorkflowParameters copy() {
     SignalExternalWorkflowParameters result = new SignalExternalWorkflowParameters();
     result.setInput(input);
     result.setRunId(runId);
+    result.setDomain(domain);
```
davidporter-id-au, line 121: I guess this was missing also?

shijiesheng (approved):

PR #948, opened 2024-11-06
fix bug in QueryWorkflowParameters.toString

shijiesheng:
**What changed?**
* Fixed a bug that toString will error if input length is smaller than 512
* remove unused method (with method has been replaced with setter already)
* add test coverage

**Why?**
part of unit test coverage

**How did you test it?**
unit test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/src/main/java/com/uber/cadence/internal/replay/QueryWorkflowParameters.java
+++ b/src/main/java/com/uber/cadence/internal/replay/QueryWorkflowParameters.java
@@ -45,22 +45,12 @@ public class QueryWorkflowParameters implements Cloneable {
     this.input = input;
   }
 
-  public QueryWorkflowParameters withInput(byte[] input) {
-    this.input = input;
-    return this;
-  }
-
   public String getRunId() {
     return runId;
   }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/uber/cadence/internal/replay/QueryWorkflowParameters.java
```diff
@@ -45,11 +45,6 @@ public void setInput(byte[] input) {
     this.input = input;
   }
 
-  public QueryWorkflowParameters withInput(byte[] input) {
```
davidporter-id-au, line 48: I don't feel strongly, but isn't the `withX` pattern preferable to setters because it's chainable?

davidporter-id-au (approved):

src/main/java/com/uber/cadence/internal/replay/QueryWorkflowParameters.java
shijiesheng: right but they are currently unused though

PR #950, opened 2024-11-07
Fix Error handling and add test coverage for Thrift2ProtoAdapter

natemort:
**What changed?**
- Remove request logging from GrpcStubs. This was a debugging tool added during implementation and isn't generally useful.
- Fix GRPC error handling for the handful of Async methods. These were previously being completed with an ExecutionException wrapping the StatusRuntimException rather than the StatusRuntimeException being mapped to the appropriate type.
- Increase test coverage

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/src/main/java/com/uber/cadence/internal/compatibility/Thrift2ProtoAdapter.java
+++ b/src/main/java/com/uber/cadence/internal/compatibility/Thrift2ProtoAdapter.java
@@ -109,6 +109,7 @@ import io.grpc.Deadline;
 import io.grpc.StatusRuntimeException;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.TimeUnit;
 import org.apache.thrift.TException;
@@ -203,7 +204,7 @@ public class Thrift2ProtoAdapter implements IWorkflowService {
       RestartWorkflowExecutionRequest restartRequest)
       throws BadRequestError, ServiceBusyError, DomainNotActiveError, LimitExceededError,
           EntityNotExistsError, ClientVersionNotSupportedError, TException {
-    throw new IllegalArgumentException("unimplemented");
+    throw new UnsupportedOperationException("unimplemented");
   }
 
   @Override
@@ -851,7 +852,7 @@ public class Thrift2ProtoAdapter implements IWorkflowService {
   public void RestartWorkflowExecution(
       RestartWorkflowExecutionRequest restartRequest, AsyncMethodCallback resultHandler)
       throws TException {
-    throw new IllegalArgumentException("unimplemented");
+    throw new UnsupportedOperationException("unimplemented");
   }
 
   @Override
@@ -880,7 +881,7 @@ public class Thrift2ProtoAdapter implements IWorkflowService {
               resultHandler.onComplete(
                   ResponseMapper.startWorkflowExecutionAsyncResponse(response));
             } catch (Exception e) {
-              resultHandler.onError(e);
+              handleAsyncException(resultHandler, e);
             }
           },
           ForkJoinPool.commonPool());
@@ -1003,7 +1004,7 @@ public class Thrift2ProtoAdapter implements IWorkflowService {
               com.uber.cadence.api.v1.SignalWorkflowExecutionResponse response = resultFuture.get();
               resultHandler.onComplete(null);
             } catch (Exception e) {
-              resultHandler.onError(e);
+              handleAsyncException(resultHandler, e);
             }
           },
           ForkJoinPool.commonPool());
@@ -1025,7 +1026,7 @@ public class Thrift2ProtoAdapter implements IWorkflowService {
       SignalWithStartWorkflowExecutionAsyncRequest signalWithStartRequest,
       AsyncMethodCallback resultHandler)
       throws TException {
-    throw new IllegalArgumentException("unimplemented");
+    throw new UnsupportedOperationException("unimplemented");
   }
 
   @Override
@@ -1199,7 +1200,7 @@ public class Thrift2ProtoAdapter implements IWorkflowService {
               com.uber.cadence.api.v1.StartWorkflowExecutionResponse response = resultFuture.get();
               resultHandler.onComplete(ResponseMapper.startWorkflowExecutionResponse(response));
             } catch (Exception e) {
-              resultHandler.onError(e);
+              handleAsyncException(resultHandler, e);
             }
           },
           ForkJoinPool.commonPool());
@@ -1230,7 +1231,7 @@ public class Thrift2ProtoAdapter implements IWorkflowService {
               resultHandler.onComplete(
                   ResponseMapper.startWorkflowExecutionAsyncResponse(response));
             } catch (Exception e) {
-              resultHandler.onError(e);
+              handleAsyncException(resultHandler, e);
             }
           },
           ForkJoinPool.commonPool());
@@ -1276,7 +1277,7 @@ public class Thrift2ProtoAdapter implements IWorkflowService {
               resultHandler.onComplete(
                   ResponseMapper.getWorkflowExecutionHistoryResponse(response));
             } catch (Exception e) {
-              resultHandler.onError(e);
+              handleAsyncException(resultHandler, e);
             }
           },
           ForkJoinPool.commonPool());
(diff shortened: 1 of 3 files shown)


src/main/java/com/uber/cadence/internal/compatibility/Thrift2ProtoAdapter.java
```diff
@@ -1293,4 +1294,13 @@ public void SignalWorkflowExecutionWithTimeout(
       throws TException {
     throw new UnsupportedOperationException("not implemented");
   }
+
+  private void handleException(AsyncMethodCallback callback, Exception exception) {
+    if (exception instanceof ExecutionException
```
shijiesheng, line 1299: This behavior is different from the non-async methods. They throw `StatusRuntimeException` directly. Could you point me where we return `ExecutionException` wrapped exceptions?

src/main/java/com/uber/cadence/internal/compatibility/proto/serviceclient/GrpcServiceStubs.java
```diff
@@ -124,9 +123,6 @@ final class GrpcServiceStubs implements IGrpcServiceStubs {
             MetadataUtils.newAttachHeadersInterceptor(headers),
             newOpenTelemetryInterceptor(),
             newOpenTracingInterceptor(options.getTracer()));
-    if (log.isTraceEnabled()) {
-      interceptedChannel = ClientInterceptors.intercept(interceptedChannel, tracingInterceptor);
-    }
```
shijiesheng, line 129: I would probably just keep it. For unit test, we can just cover the lines without asserting as it's for debugging purposes.

shijiesheng: Also seeing this pattern in multiple other entities. For example ActivityPollTask

src/main/java/com/uber/cadence/internal/compatibility/Thrift2ProtoAdapter.java
natemort: Within the async methods we have blocks that look like this:
```
      resultFuture.addListener(
          () -> {
            try {
              com.uber.cadence.api.v1.GetWorkflowExecutionHistoryResponse response =
                  resultFuture.get();
              resultHandler.onComplete(
                  ResponseMapper.getWorkflowExecutionHistoryResponse(response));
            } catch (Exception e) {
              resultHandler.onError(e);
            }
          }
```

If the `resultFuture` [completed with an exception](https://docs.oracle.com/javase/8/docs/api/index.html?java/util/concurrent/Future.html) it will throw an ExecutionException wrapping the actual cause of the failure (the StatusRuntimeException). With our current implementation this has two issues:
- We're completing the resultHandler with the ExecutionException.
- We don't map the StatusRuntimeException to the thrift error type.

So currently if a user should receive an AccessDeniedError they'll receive an ExecutionException containing a StatusRuntimeException.

shijiesheng: got it. Maybe we can rename it as `handleAsyncException`?

src/main/java/com/uber/cadence/internal/compatibility/proto/serviceclient/GrpcServiceStubs.java
natemort: Sounds good, readded with test coverage.

src/main/java/com/uber/cadence/internal/compatibility/Thrift2ProtoAdapter.java
natemort: Sounds good, done.

src/main/java/com/uber/cadence/internal/compatibility/proto/serviceclient/GrpcServiceStubs.java
```diff
@@ -124,7 +127,7 @@ final class GrpcServiceStubs implements IGrpcServiceStubs {
             MetadataUtils.newAttachHeadersInterceptor(headers),
             newOpenTelemetryInterceptor(),
             newOpenTracingInterceptor(options.getTracer()));
-    if (log.isTraceEnabled()) {
+    if (log.isTraceEnabled() || enableLogging) {
```
shijiesheng, line 130: chatgpt says you can change logger level in the code directly. If that's true, we may not need `enableLogging`
```
@Before
    public void setUp() {
        // Set the logging level to TRACE for detailed output in tests
        LoggerContext loggerContext = (LoggerContext) LoggerFactory.getILoggerFactory();
        loggerContext.getLogger("ROOT").setLevel(Level.TRACE);
    }
```

src/main/java/com/uber/cadence/internal/compatibility/proto/serviceclient/GrpcServiceStubs.java
```diff
@@ -91,7 +93,8 @@ final class GrpcServiceStubs implements IGrpcServiceStubs {
   private final MetaAPIGrpc.MetaAPIBlockingStub metaBlockingStub;
   private final MetaAPIGrpc.MetaAPIFutureStub metaFutureStub;
 
-  GrpcServiceStubs(ClientOptions options) {
+  @VisibleForTesting
+  public GrpcServiceStubs(ClientOptions options, boolean enableLogging) {
```
shijiesheng, line 97: instead of making it public, you could use the newInstance method https://github.com/uber/cadence-java-client/blob/bfd521dc895a59ffb2ba17875f828072e36a64b1/src/main/java/com/uber/cadence/internal/compatibility/proto/serviceclient/IGrpcServiceStubs.java#L38

src/main/java/com/uber/cadence/internal/compatibility/proto/serviceclient/GrpcServiceStubs.java
```diff
@@ -53,7 +54,8 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-final class GrpcServiceStubs implements IGrpcServiceStubs {
+@VisibleForTesting
+public final class GrpcServiceStubs implements IGrpcServiceStubs {
```
shijiesheng, line 58: same as below, it doesn't need to public

src/test/java/com/uber/cadence/internal/compatibility/ProtoObjects.java
```diff
@@ -1324,6 +1324,65 @@ public final class ProtoObjects {
...
+
+  public static final ResetStickyTaskListResponse RESET_STICKY_TASK_LIST_RESPONSE =
+      ResetStickyTaskListResponse.getDefaultInstance();
+
+  public static final RefreshWorkflowTasksRequest REFRESH_WORKFLOW_TASKS_REQUEST =
+      RefreshWorkflowTasksRequest.getDefaultInstance();
+
+  public static final RefreshWorkflowTasksResponse REFRESH_WORKFLOW_TASKS_RESPONSE =
+      RefreshWorkflowTasksResponse.getDefaultInstance();
+
```
davidporter-id-au, line 1385: I'm not 100% sure how much we care about these, but if we do care about them, is there some kind of assertion we can add to ensure that they're mapped for all future fields too? (maybe reflection?)

src/main/java/com/uber/cadence/internal/compatibility/proto/serviceclient/GrpcServiceStubs.java
natemort: I initially made it public to pass through the enableLogging field without adding it to IGrpcServiceStubs#newInstance, but using the approach you've suggested I've enabled tracing and removed that field. Done.

src/main/java/com/uber/cadence/internal/compatibility/proto/serviceclient/GrpcServiceStubs.java
natemort: Done.

src/main/java/com/uber/cadence/internal/compatibility/proto/serviceclient/GrpcServiceStubs.java
natemort: Done

src/test/java/com/uber/cadence/internal/compatibility/ProtoObjects.java
natemort: For any type that we have a mapper for we use [MapperTestUtil](https://github.com/uber/cadence-java-client/blob/master/src/test/java/com/uber/cadence/internal/compatibility/MapperTestUtil.java) to assert that all fields are set (or a select set are missing if we haven't updated the mappers for them). We do similar things with some of the types contained within the requests(decisions and history events) as well as enums.

These types don't actually exist in thrift (the thrift APIs return void, something we never do in gRPC) so there aren't any mappers. It seems very unlikely we'd add fields to these types until after we completely remove thrift, at which point this will Thrift2Proto layer won't exist at all. So I don't think it's worth testing these ones the way all the others are tested.

shijiesheng (approved):

PR #978, opened 2024-11-28
Added tests for LocalActivityOptions

jakobht:
**What changed?**
Added tests for LocalActivityOptions so we are now at 89% coverage for this file

**Why?**
Increase coverage

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/src/test/java/com/uber/cadence/activity/LocalActivityOptionsTest.java
+++ b/src/test/java/com/uber/cadence/activity/LocalActivityOptionsTest.java
@@ -1,0 +1,75 @@
+/*
+ *
+ *  *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  *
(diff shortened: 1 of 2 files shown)


src/test/java/com/uber/cadence/activity/LocalActivityOptionsTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+public class LocalActivityOptionsTest extends TestCase {
+
+  @Test
+  public void testTestToString() {
+    String asString = createOptions().toString();
+
+    final String expected =
+        "LocalActivityOptions{"
+            + "scheduleToCloseTimeout=PT2M3S, "
```
dkrotx, line 37: I would expect it to be more human-readable, but that's OK as well :-)

dkrotx (approved):

PR #979, opened 2024-11-28
Test for TestWorkflowMutableStateImpl

jakobht:
**What changed?**
- Added tests for TestWorkflowMutableStateImpl
- Detracted validation methods to static functions as they are not using the instance variables anyway
- Split one validation method in two so one validates and the other mutates 

**Why?**
Increase code coverage

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


3vilhamster (approved):
Good job! It is not only unit tests but also improving code structure/mix of responsibilities and weird internalization.

PR #983, opened 2025-02-25
Update README.md

ibarrajo:
fix typo and references from `uber/cadence` to `cadence-workflow/cadence`


shijiesheng:
addressed here https://github.com/cadence-workflow/cadence-java-client/pull/1053

PR #988, opened 2025-03-26
added ExecutorWrapper in WorkerFactoryOptions

shijiesheng:
**What changed?**

* added new option in WorkerFactoryOptions to expose user provided ExecutorWrapper
* added a ExecutorWrapper interface
* apply ExecutorWrapper to executors in heartbeat, pollers (both poller executor and poller task executor), SyncActivityWorker and SyncWorkflowWorker. 

**Why?**

By wrapping the ExecutorWrapper, it is now possible to propagate "root" context in background polling and task executions. Uber internally built a threadlocal storage to propagate context inside executors. 

**How did you test it?**

Unit Test
WIP internal test samples on context propagation

**Potential risks**

No risk due to it's a new fiel with fallback on original behavior

**Release notes**

**Documentation Changes**


natemort (approved):
It would be nice to have tests for this but I can't think of any reasonable way to go about that.

PR #989, opened 2025-04-08
add serviceclient exceptions

shijiesheng:
**What changed?**

* added all exceptions to match thrift exceptions
* added ErrorMapper which maps StatusRuntimeException to ServiceClientException

**Why?**

thrift exceptions are used across Java SDK. It's easier to abstract the protocol level error handling by introducing internal exception layer

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


src/main/java/com/uber/cadence/serviceclient/exceptions/QueryFailedException.java
```diff
@@ -0,0 +1,4 @@
+package com.uber.cadence.serviceclient.exceptions;
+
+public class QueryFailedException extends ServiceClientException {
```
natemort, line 3: This is unused

src/main/java/com/uber/cadence/serviceclient/exceptions/StickyWorkerUnavailableException.java
```diff
@@ -0,0 +1,4 @@
+package com.uber.cadence.serviceclient.exceptions;
+
+public class StickyWorkerUnavailableException extends ServiceClientException {
```
natemort, line 3: This is unused and I don't think can actually propagate to the client

src/main/java/com/uber/cadence/internal/compatibility/proto/ErrorMapper.java
```diff
@@ -0,0 +1,70 @@
...
+import com.uber.cadence.serviceclient.exceptions.*;
+import io.grpc.StatusRuntimeException;
+import io.grpc.protobuf.StatusProto;
+
+public class ErrorMapper {
+    public static ServiceClientException Error(StatusRuntimeException e) {
+
+        Status status = StatusProto.fromThrowable(e);
+        if (status == null || status.getDetailsCount() == 0) {
+            return new ServiceClientException("empty status or status with empty details", e);
```
natemort, line 17: Nit: When can this happen? What kind of exception does GRPC throw if we have a connection issue rather than an error returned by the server?

natemort (approved):

src/main/java/com/uber/cadence/internal/compatibility/proto/ErrorMapper.java
shijiesheng: Server controls whether to pass details. https://github.com/uber/cadence/blob/f4e219ae4248552e53a8ba57036b9af62485221b/common/types/mapper/proto/errors.go#L274
I've simplified the logic to handle such cases.

src/main/java/com/uber/cadence/serviceclient/exceptions/QueryFailedException.java
shijiesheng: deleted

src/main/java/com/uber/cadence/serviceclient/exceptions/StickyWorkerUnavailableException.java
shijiesheng: deleted

PR #997, opened 2025-05-06
Change license to Apache 2.0

demirkayaender:
**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "src/main/idls"]
 	path = src/main/idls
-	url = git@github.com:uber/cadence-idl.git
+	url = https://github.com/cadence-workflow/cadence-idl.git
 	branch = master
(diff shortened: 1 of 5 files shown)


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "src/main/idls"]
 	path = src/main/idls
-	url = git@github.com:uber/cadence-idl.git
```
davidporter-id-au, line 3: did you mean to change this? This should be `git@` as far as I'm aware

demirkayaender: Is that a java thing? Why should it be git@? All the other ones are using this format and it seems to be working.
This also seems to be working to me.

davidporter-id-au: it refs to the git cloning option, by default git uses SSH as a transport protocol, but you can use HTTP as well, though it's a bit less well supported. 
Screenshot 2025-05-05 at 7 54 13 PM

shijiesheng (approved):

PR #999, opened 2025-05-15
added WorkflowServiceGrpc and IWorkflowServiceV4

shijiesheng:
**What changed?**

* added client entities that maps 1:1 with thrift (generated from a generator)
* added `BaseError`, which is a catchall similar to `TException`. 
* added mappers to transform Proto to client entities. (mostly copied from thrift mappers)
* added IWorkflowServiceV4 which is exactly same as IWorkflowService except for it's using the client entities. 
* added WorkflowServiceGrpc that implements the V4 interface

**Plan**

* new V4 interface with new entities < This PR
* replace all pointers to V4 in one go (Should be import change only)

**Why?**

Thrift deprecation work

**How did you test it?**

Unit Test on mappers from ClientObjects to ProtoObjects

**Potential risks**

**Release notes**

**Documentation Changes**


shijiesheng:
> Is this a complete switch to GRPC, so we're throwing away thrift? I wonder how much of the diff is auto-generated and with one worth checking? the diff looks rather huge.

I finally get the chance to add the unit test for mappers. 

To answer your question, I mainly need reviews on 1:1 comparison of thrift vs newly added entities. Others are captured by mappers tests.

build.gradle
```diff
@@ -89,6 +89,11 @@ dependencies {
     compile group: 'com.google.api.grpc', name: 'proto-google-common-protos', version: '2.10.0'
     compile group: 'com.google.protobuf', name: 'protobuf-java-util', version: '3.21.9'
     compile group: 'com.google.oauth-client', name: 'google-oauth-client', version: '1.35.0'
+    compile group: 'org.projectlombok', name: 'lombok', version: '1.18.30'
```
natemort, line 92: We should only need the compileOnly dependency. We don't want to include this at runtime.

timl3136 (approved):

src/main/java/com/uber/cadence/internal/compatibility/proto/mappers/DecisionMapper.java
```diff
@@ -0,0 +1,274 @@
...
+import com.uber.cadence.api.v1.StartChildWorkflowExecutionDecisionAttributes;
+import com.uber.cadence.api.v1.StartTimerDecisionAttributes;
+import com.uber.cadence.api.v1.UpsertWorkflowSearchAttributesDecisionAttributes;
+import java.util.ArrayList;
+import java.util.List;
+
+class DecisionMapper {
+  static List<Decision> decisionArray(List<com.uber.cadence.entities.Decision> t) {
+    if (t == null) {
+      return null;
```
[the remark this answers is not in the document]
shijiesheng: fixed

build.gradle
```diff
@@ -139,6 +143,7 @@ sourceSets {
         }
         java {
             srcDir 'src/main'
+            srcDir '.gen'
```
natemort, line 146: With idiomatic Java you pretty much never check in generated sources (unlike Go).

But we don't want to run the generator as part of the build step and want to eventually delete the generator entirely. At which point we'd want to put it in src/main/java with all the other sources.

I think it makes the most sense to put it in like `src/gen/java/` and then we merge it into `src/main/java` later.

The `.gen` directory pattern just seems really strange from a Java perspective.

natemort (approved):

build.gradle
shijiesheng: fixed

PR #1001, opened 2025-05-20
support for "test runner for java" extension in vscode

shijiesheng:
**What changed?**

* disable ErrorProne checks (some files are already violating but is suppressed due to Xlint:none)

**Why?**

Test Runner for Java extension for some reason doesn't read Xlint:none flag and thus fails on suppressed lint errors

**How did you test it?**

Locally tested on vscode

Before
Screenshot 2025-05-20 at 12 47 32 PM

After
Screenshot 2025-05-20 at 12 48 56 PM

**Potential risks**

**Release notes**

**Documentation Changes**


natemort (approved):
Approved, once we're in v 4.x we should be able to reintroduce these and upgrade the gradle version

Issue #998, opened 2025-05-13
Workflow cancellation cause is not captured in history


Ramkumar92:
While cancelling the workflow using cadence-java-client, the cause is not captured in history
    
    RequestCancelWorkflowExecutionRequest requestCancelWorkflowExecutionRequest =
        new RequestCancelWorkflowExecutionRequest()
            .setDomain(domain)
            .setWorkflowExecution(new WorkflowExecution().setWorkflowId(workflowId))
            .setCause('Cancellation reason');
    workflowClient
        .getService()
        .RequestCancelWorkflowExecution(requestCancelWorkflowExecutionRequest);

When the cancellation is done from cadence web UI the cause is defaulted and visible in workflow history, so this appears to be a bug with cadence-java-client.

This appears to be due the cause not being set in [RequestMapper.java](https://github.com/cadence-workflow/cadence-java-client/blob/aaf9d1d522b91d7915e80ac3bb669bf5faa19c23/src/main/java/com/uber/cadence/internal/compatibility/proto/RequestMapper.java#L171)

PR #1055, opened 2026-05-01
feat: Bring Pollers & Workers on par with Go implementations #998

Shaddoll:
**What changed?**

Cherry pick several changes from temporal to cadence

relates to https://github.com/temporalio/sdk-java/issues/998

**Why?**
See https://github.com/temporalio/sdk-java/issues/998.

Another reason is that we want to enforce rate limiting on sticky pollers. 

**How did you test it?**
./gradlew test

**Potential risks**

This changes several components in java client. It could make workflow/poller to be stuck if there is anything wrong.

**Release notes**

**Documentation Changes**


src/main/java/com/uber/cadence/internal/worker/WorkflowWorker.java
```diff
@@ -171,12 +190,36 @@ private byte[] queryWorkflowExecution(
   }
 
   @Override
-  public void accept(PollForDecisionTaskResponse pollForDecisionTaskResponse) {
-    pollTaskExecutor.process(pollForDecisionTaskResponse);
+  public void shutdown() {
```
natemort: Nit: no need for these overrides that just call super.

src/main/java/com/uber/cadence/worker/WorkerFactoryOptions.java
```diff
@@ -87,18 +81,22 @@ public Builder setMaxWorkflowThreadCount(int maxWorkflowThreadCount) {
     /**
      * Timeout for sticky workflow decision to be picked up by the host assigned to it. Once it
      * times out then it can be picked up by any worker. Default value is 5 seconds.
+     *
+     * @deprecated use {@link
+     *     WorkerOptions.Builder#setStickyTaskListScheduleToStartTimeout(Duration)} to specify this
+     *     value per-worker instead
      */
+    @Deprecated
     public Builder setStickyTaskScheduleToStartTimeout(Duration stickyTaskScheduleToStartTimeout) {
```
natemort: How much effort is it to still support these properties? Alternatively, maybe we should just remove them so that it's much more obvious that we introduced a behavioral change here.

I don't see usage of either within Uber so this would be more impactful on our open source users.

src/main/java/com/uber/cadence/worker/WorkerOptions.java
```diff
@@ -151,7 +156,21 @@ public Builder setTracer(Tracer tracer) {
...
+     * out then it can be picked up by any worker. Default value is 5 seconds.
+     */
+    public Builder setStickyTaskListScheduleToStartTimeout(
+        Duration stickyTaskListScheduleToStartTimeout) {
+      this.stickyTaskListScheduleToStartTimeout = stickyTaskListScheduleToStartTimeout;
+      return this;
+    }
+
     public WorkerOptions build() {
+      Duration stickyTimeout = stickyTaskListScheduleToStartTimeout;
```
natemort: Nit: do this in the setter

src/main/java/com/uber/cadence/internal/worker/StickyQueueBalancer.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  /** @return task list kind that should be used for the next poll */
+  public synchronized TaskListKind makePoll() {
+    if (stickyQueueEnabled) {
+      // If pollersCount >= stickyBacklogSize > 0 we want to go back to a normal ratio to avoid a
+      // situation that too many pollers (all of them in the worst case) will open only sticky queue
+      // polls observing a stickyBacklogSize == 1 for example (which actually can be 0 already at
+      // that moment) and get stuck causing dip in worker load.
+      if (stickyBacklogSize > pollersCount || stickyPollers <= normalPollers) {
```
natemort, line 42: I'm surprised that we don't guarantee at least one Poll for each Kind. That seems like a safer/more conservative behavior.

src/main/java/com/uber/cadence/internal/worker/WorkflowPollTask.java
```diff
@@ -31,104 +31,163 @@
...
-            metricScope, MetricsType.DECISION_POLL_LATENCY, HistogramBuckets.DEFAULT_1MS_100S);
+    this.stickyMetricScope =
+        metricScope.tagged(
+            new ImmutableMap.Builder<String, String>(1)
+                .put(MetricsTag.TASK_LIST, String.format("%s:%s", taskList, "sticky"))
+                .build());
 
-    PollForDecisionTaskRequest pollRequest = new PollForDecisionTaskRequest();
+    // Normal poll request
+    this.pollRequest = new PollForDecisionTaskRequest();
```
natemort: I don't think we should reused request objects. Thrift types are mutable and we don't know what interceptors or wrappers have been added to the IWorkflowService. It would be safe if these were proto objects, but alas.

As an example, our internal test interceptor updates the request objects to rewrite the TaskList instead of copying it to a new object. It doesn't have a guard against re-encoding the same TaskList, so it would probably break with this change.

We could just duplicate these objects before every request

src/main/java/com/uber/cadence/internal/worker/WorkflowPollTask.java
```diff
@@ -31,104 +31,163 @@
...
-    TaskList tl = new TaskList().setName(taskList).setKind(taskListKind);
+    TaskList tl = new TaskList().setName(taskList).setKind(TaskListKind.NORMAL);
     pollRequest.setTaskList(tl);
 
-    if (log.isDebugEnabled()) {
-      log.debug("poll request begin: " + pollRequest);
-    }
-    PollForDecisionTaskResponse result;
+    // Sticky poll request
+    this.stickyPollRequest = new PollForDecisionTaskRequest();
```
natemort: Nit: copy the above request and just change the TaskList to avoid the initialization logic ever drifting.

src/main/java/com/uber/cadence/worker/Worker.java
```diff
@@ -142,11 +147,17 @@ public final class Worker implements Suspendable {
...
-            stickyDecisionScheduleToStartTimeout,
+            enableStickyExecution ? getStickyTaskListName(client.getOptions().getIdentity()) : null,
             threadPoolExecutor);
   }
 
+  private static String getStickyTaskListName(String workerIdentity) {
+    // Unique id is needed to avoid collisions with other workers that may be created for the same
+    // task list and with the same identity.
+    UUID uniqueId = UUID.randomUUID();
+    return String.format("%s:%s", workerIdentity, uniqueId);
```
natemort, line 158: This looks slightly different from the previous TaskList naming, does it match the Go client?

src/test/java/com/uber/cadence/internal/tracing/StartWorkflowTest.java
```diff
@@ -177,7 +177,8 @@ public void testStartMultipleWorkflowGRPC() {
     Worker worker;
     worker =
         workerFactory.newWorker(
-            TASK_LIST, WorkerOptions.newBuilder().setMaxConcurrentWorkflowExecutionSize(2).build());
+            TASK_LIST,
+            WorkerOptions.newBuilder().setMaxConcurrentWorkflowExecutionSize(20).build());
```
natemort, line 181: Nit: Is this still needed?

src/main/java/com/uber/cadence/worker/WorkerFactoryOptions.java
Shaddoll: I don't want to make this change a breaking change.

I think the effort is not much.

src/main/java/com/uber/cadence/worker/Worker.java
Shaddoll: It's a bit different from go client because it also includes the pid

src/test/java/com/uber/cadence/internal/tracing/StartWorkflowTest.java
Shaddoll: the concurrent size needs to be larger than 2

src/main/java/com/uber/cadence/worker/WorkerFactoryOptions.java
natemort: It's breaking change either way, the difference is whether it's a breaking change at compile time or a breaking change at runtime due to behavioral differences.

I think users would generally be better off by the breaking change at compile time that gives them a chance to update their code and ensure it continues to behave the same. Particularly for the sticky schedule to start timeout, they may rely on setting very low values.

Since we don't use these methods at all within Uber there's no additional effort for us. For external users, if they rely on these properties then they need to update their code anyways.

Shaddoll: makes sense!

src/test/java/com/uber/cadence/internal/tracing/StartWorkflowTest.java
natemort: But why? Shouldn't we be evicting sticky workflows for incoming workflows? This shouldn't impact it

src/main/java/com/uber/cadence/internal/worker/WorkflowPollTask.java
```diff
@@ -31,104 +31,160 @@
...
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      return null;
+    }
 
-    TaskList tl = new TaskList().setName(taskList).setKind(taskListKind);
-    pollRequest.setTaskList(tl);
+    TaskListKind taskListKind = stickyQueueBalancer.makePoll();
+    boolean isSticky = TaskListKind.STICKY.equals(taskListKind);
+    PollForDecisionTaskRequest request =
```
natemort, line 88: Nit: move to separate method, pass the TaskList as a parameter

natemort (approved):

PR #1060, opened 2026-06-02
fix: Data race on failedPromises in DeterministicRunnerImpl

hcantunc:
**What changed?**
 Replace `HashSet` with `ConcurrentHashMap.newKeySet()` for `failedPromises` in `DeterministicRunnerImpl`.

**Why?**
`failedPromises` is written to via `registerFailedPromise`/`forgetFailedPromise` without holding the lock, while `close()` reads it under the lock on another thread. 

**How did you test it?**
Existing unit tests.

**Potential risks**
Minimal. `ConcurrentHashMap.newKeySet()` is a drop-in replacement for `HashSet` but has slightly higher memory overhead and per-operation.

**Release notes**
N/A

**Documentation Changes**
N/A

### DIFF
--- a/src/main/java/com/uber/cadence/internal/sync/DeterministicRunnerImpl.java
+++ b/src/main/java/com/uber/cadence/internal/sync/DeterministicRunnerImpl.java
@@ -53,6 +53,7 @@ import java.util.Optional;
 import java.util.Random;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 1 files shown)


bowenxia:
Do we have a unit test to cover the Data race on failedPromises case?

shijiesheng:
It's not needed because DeterministicRunner is single threaded by design. (A coroutine)

shijiesheng (approved):
Thank you for the change. Confirmed that we do have two threads accessing the failedPromises. (workflow thread and decision handler thread)

PR #1067, opened 2026-06-30
feat(schedules): add schedule methods to IWorkflowService

abhishekj720:
## What changed?

Adds the client-facing schedule model types and the schedule service contract.

**Model types** (new package `com.uber.cadence.client.schedule`):
- `ScheduleSpec`: cron expression, start/end window, jitter
- `ScheduleAction` / `StartWorkflowAction`: what to run on each fire
- `SchedulePolicies`: overlap, catch-up, buffer/concurrency limits, pause-on-failure
- `ScheduleOverlapPolicy`, `ScheduleCatchUpPolicy`: clean client-side enums (no Thrift `INVALID` sentinel)
- `ScheduleState`, `ScheduleInfo`: immutable snapshots returned by `describeSchedule`
- `ScheduleDescription`: immutable full description returned by `describeSchedule`; maps are defensively copied and guaranteed non-null
- `ScheduleListEntry`, `ListSchedulesResult`: paged list response from `listSchedules`

**`ScheduleBackfill`** (in `com.uber.cadence.client`) use `Instant` for start/end time, consistent with Thrift epoch-nano wire format and `ScheduleSpec`; adds `equals()`/`hashCode()` for value semantics

**Service contract**: 8 schedule operations added to `IWorkflowService`: `createSchedule`, `describeSchedule`, `updateSchedule`, `deleteSchedule`, `pauseSchedule`, `unpauseSchedule`, `backfillSchedule`, `listSchedules`

`UnsupportedOperationException` stubs provided in `IWorkflowServiceBase`, `WorkflowServiceGrpc`, `TestWorkflowService`, and the two test-environment wrapper classes; all stubs follow the existing pattern in each class.

## Why?
Downstream PRs (gRPC implementation, `WorkflowClient` methods) need stable model types and a service interface to implement against. Adding both here lets each layer be reviewed independently.

## How did you test it?
- `./gradlew compileJava` all implementors satisfy the contract via the `IWorkflowServiceBase` defaults
- `./gradlew verifyGoogleJavaFormat` formatting clean
- Unit tests in `ScheduleTypesTest` cover: builders, getters, immutability/defensive-copy, equals/hashCode, toString for all model types

## Potential risks
None; purely additive. All new service methods throw `UnsupportedOperationException` until the gRPC implementation lands.

## Release notes
N/A internal model types and service interface; no user-visible change until client methods land in a follow-up PR.

## Documentation Changes
None.


src/main/java/com/uber/cadence/client/schedule/ListSchedulesResult.java
```diff
@@ -0,0 +1,88 @@
+/**
```
shijiesheng: We don't need these entities. Can we use gen entities directly? We don't need ScheduleBackfill as well because we can directly use BackfillScheduleRequest

src/main/java/com/uber/cadence/client/ScheduleClient.java
shijiesheng: I would still want to pass the response object in case we add fields later and then it's not a breaking change. Overload doesn't allow response type changes.

src/main/java/com/uber/cadence/client/schedule/ListSchedulesResult.java
abhishekj720: Deleted ListSchedule. But kept ScheduleBackfill because BackfillScheduleRequest has domain, scheduleId, startTimeNano (raw long), and a Thrift enumm all things callers shouldn't manage directly.

src/main/java/com/uber/cadence/client/ScheduleClient.java
abhishekj720: Passed

src/main/java/com/uber/cadence/serviceclient/IWorkflowService.java
```diff
@@ -816,4 +816,89 @@ void ListTaskListPartitions(
...
+
+  /**
+   * Creates a new schedule in the domain.
+   *
+   * @throws BadRequestError if the schedule spec is invalid (e.g. bad cron expression)
+   * @throws WorkflowExecutionAlreadyStartedError if a schedule with the same ID already exists
+   * @throws EntityNotExistsError if the domain does not exist
+   * @throws CadenceError on other server errors
+   */
+  CreateScheduleResponse CreateSchedule(CreateScheduleRequest request)
```
natemort, line 828: Nit: You probably want to add the CompletableFuture variants of these to match the blocking ones, since ScheduleClient is async.

natemort (approved):

src/main/java/com/uber/cadence/serviceclient/IWorkflowService.java
abhishekj720: Will fix this in a follow up PR, thanks

PR #1074, opened 2026-08-12
feat(ContextPropagator): Replace imperative set/unset context propagation with lexically-scoped runWithContext

rlei:
## Summary
- Adds `ContextPropagator.runWithContext(Object context, ContextRunnable task)`, a default method that pairs `setCurrentContext`/`task.run()`/`unsetCurrentContext` in a single try/finally, replacing the previous pattern where context was set at the top of a handler and unset in a separate `finally` block possibly several methods away.
- `ContextThreadLocal.runWithContext(...)` composes all configured propagators into a single nested call chain executed lexically around the task.
- Updates `ActivityWorker`, `LocalActivityWorker`, and `WorkflowThreadImpl` to run their task through this single call.

## Why
JDK 25 finalizes `ScopedValue` (JEP 481), a lexically-scoped, immutable alternative to `ThreadLocal`: a value is bound and the task runs inside that binding in a single call (`ScopedValue.where(...).call(task)`), with no separate unset step. The previous `ContextPropagator` API couldn't express that shape. Restructuring around a single `runWithContext(context, task)` call:
- **Simplicity**: a `ScopedValue`-backed propagator only needs to override `runWithContext` and call `ScopedValue.where(VALUE, context).run(task)` — no separate set/get/unset methods to keep consistent with each other.
- **Immutable context data**: `ScopedValue` bindings are write-once for the dynamic extent of the call; nothing downstream can accidentally mutate a shared context object the way a mutable `ThreadLocal` value can be.
- **No unpaired unset**: because the binding and the task execution happen in the same call frame, cleanup is guaranteed by stack unwinding — removing the risk of an early return, a thrown exception, or a future refactor leaving a propagator's context set with no matching unset.

Existing propagators that implement `setCurrentContext`/`unsetCurrentContext` are unaffected — `runWithContext` has a default implementation built on those methods, so this is additive for current implementations and opt-in for `ScopedValue`-based ones.

## Test plan
- [x] `./gradlew test` (full suite) passes
- [x] New `ContextThreadLocalTest` covers nested propagator ordering, lexical-scope override behavior, and unset-after-failure behavior


src/main/java/com/uber/cadence/context/ContextPropagator.java
```diff
@@ -139,4 +139,27 @@ public interface ContextPropagator {
...
+
+  /**
+   * Executes {@code task} with {@code context} installed as the current context.
+   *
+   * <p>The default implementation preserves the legacy imperative context lifecycle. Propagators
+   * that use lexical context, such as Java Scoped Values, should override this method and execute
+   * {@code task} inside their lexical binding rather than implementing {@link #setCurrentContext}
+   * and {@link #unsetCurrentContext}.
+   */
+  default void runWithContext(Object context, ContextRunnable task) throws Exception {
```
shijiesheng, line 167: use callable to save one additional `ContextRunnable` interface.
```
default <T> T callWithContext(Object context, Callable<T> body) throws Exception {
  setCurrentContext(context);
  try {
    return body.call();
  } finally {
    unsetCurrentContext();
  }
}
```

rlei: the PR was initially created using `Callable`, but I dropped it for the simpler runnable, as there's no use cases that need a return value in Cadence. Using Callable will add a couple of unused `return null;` here and there.

src/main/java/com/uber/cadence/internal/context/ContextThreadLocal.java
```diff
@@ -57,20 +58,38 @@ public static Map<String, Object> getCurrentContextForPropagation() {
     return contextData;
   }
 
-  public static void propagateContextToCurrentThread(Map<String, Object> contextData) {
-    if (contextData == null || contextData.isEmpty()) {
+  public static void runWithContext(Map<String, Object> contextData, ContextRunnable task)
```
shijiesheng, line 63: From AI: custom code can now swallow exceptions that WorkflowThread throws. Do you want to add some guard on this? For example we can compare the original result, throwable and compare them with the new output. 
We also want to add some comments to discourage users to swallow and modify exceptions.

rlei: good point. discussed offline, the fix will be to detect such bad `runWithContext` implementations and fail it with a `ContextPropagatorSwallowedExceptionError` (extending `Error`) to alert the propagator owner.

src/main/java/com/uber/cadence/internal/context/ContextThreadLocal.java
```diff
@@ -57,20 +59,61 @@ public static Map<String, Object> getCurrentContextForPropagation() {
...
+          } catch (Throwable t) {
+            thrown.set(t);
+            throw t;
+          }
+        };
+    for (int i = propagators.size() - 1; i >= 0; i--) {
+      ContextPropagator propagator = propagators.get(i);
       if (contextData.containsKey(propagator.getName())) {
-        propagator.setCurrentContext(contextData.get(propagator.getName()));
+        applied.add(0, propagator);
```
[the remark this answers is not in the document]
rlei: valid concern. updated to also detect propagator retrying the task, which will also be considered a violation error.

src/main/java/com/uber/cadence/internal/context/ContextThreadLocal.java
```diff
@@ -57,20 +60,68 @@ public static Map<String, Object> getCurrentContextForPropagation() {
...
-    for (ContextPropagator propagator : contextPropagators.get()) {
-      propagator.unsetCurrentContext();
+    if (invocationCount.get() != 1) {
+      throw ContextPropagatorContractViolationError.unexpectedInvocationCount(
+          applied, invocationCount.get());
+    }
+    Throwable swallowed = thrown.get();
+    if (swallowed != null) {
+      throw ContextPropagatorContractViolationError.swallowedException(applied, swallowed);
     }
```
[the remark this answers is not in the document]
rlei: Good catch, fixing too.

shijiesheng (approved):

PR #1076, opened 2026-08-13
chore(test): fix flaky tests on TChannel span assertion

shijiesheng:
**What changed?**

* Add awaitSpan helper method
* fixed StartWorkflowTest and WorkflowClientInternalTest

**Why?**

TChannel can activate span but has race conditions on the callback. This would cause span assertions assuming it's synchronos. 

**How did you test it?**

Unit Test

**Potential risks**
no, it's pure test changes

**Release notes**

**Documentation Changes**


shijiesheng:
https://github.com/cadence-workflow/cadence-java-client/pull/1077

PR #1077, opened 2026-08-13
chore(test): fix test flakiness due to tchannel span assertion

shijiesheng:
**What changed?**

* Disabled two tests that will 100% fail the docker test because of the thrift header is now invalid to new grpc-go server. It still works today because internally we use a forked grpc library. 
* Add awaitSpan helper method
* fixed StartWorkflowTest and WorkflowClientInternalTest

**Why?**

TChannel can activate span but has race conditions on the callback. This would cause span assertions assuming it's synchronos. 

**How did you test it?**

Unit Test

**Potential risks**
no, it's pure test changes

**Release notes**

**Documentation Changes**


shijiesheng:
merged into this https://github.com/cadence-workflow/cadence-java-client/pull/1079

PR #1081, opened 2026-08-19
feat(schedule): support creating schedule in paused state

abhishekj720:
## Summary

- Adds `ScheduleInitialState`, an input-side type capturing `paused`, `pauseReason`, and `pausedBy` (`pausedAt` is excluded as it is server-populated)
- Adds a five-arg `createSchedule()` overload accepting `ScheduleInitialState`; the existing four-arg overload delegates to it with `null`, preserving backward compatibility
- Adds `toThriftInitialState()` helper in `ScheduleClientImpl` mapping `ScheduleInitialState` → `com.uber.cadence.ScheduleState`
- Updates `src/main/idls` submodule to `d6d4d81` which adds the `state` field to `CreateScheduleRequest` in `shared.thrift`
- Adds the corresponding `state` field to the generated `CreateScheduleRequest` Java class
- Adds four unit tests covering paused-with-info, paused-no-info, null-state, and four-arg-falls-through
- Companion changes: cadence server [#8431](https://github.com/cadence-workflow/cadence/pull/8431) (landed), go-client [#1521](https://github.com/cadence-workflow/cadence-go-client/pull/1521), python-client [#164](https://github.com/cadence-workflow/cadence-python-client/pull/164)

## Test plan

- [ ] `./gradlew compileJava -x generateProto`: compiles cleanly
- [ ] `./gradlew test --tests "com.uber.cadence.internal.sync.ScheduleClientImplTest"`: all tests pass (including four new `initialState` tests)

### DIFF
--- a/src/main/java/com/uber/cadence/internal/sync/ScheduleClientImpl.java
+++ b/src/main/java/com/uber/cadence/internal/sync/ScheduleClientImpl.java
@@ -45,6 +45,7 @@ import com.uber.cadence.client.schedule.ScheduleAction;
 import com.uber.cadence.client.schedule.ScheduleCatchUpPolicy;
 import com.uber.cadence.client.schedule.ScheduleDescription;
 import com.uber.cadence.client.schedule.ScheduleInfo;
+import com.uber.cadence.client.schedule.ScheduleInitialState;
 import com.uber.cadence.client.schedule.ScheduleOverlapPolicy;
 import com.uber.cadence.client.schedule.SchedulePolicies;
 import com.uber.cadence.client.schedule.ScheduleSpec;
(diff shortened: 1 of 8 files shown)


src/main/java/com/uber/cadence/internal/sync/ScheduleClientImpl.java
```diff
@@ -80,12 +81,25 @@ public CompletableFuture<CreateScheduleResponse> createSchedule(
...
+      SchedulePolicies policies,
+      ScheduleInitialState initialState) {
     try {
       CreateScheduleRequest request =
           new CreateScheduleRequest()
               .setSpec(toThriftSpec(spec))
               .setAction(toThriftAction(action))
               .setPolicies(toThriftPolicies(policies));
+      if (initialState != null) {
+        request.setState(toThriftInitialState(initialState));
```
YaweiZhang-930, line 101: We set state on the request here, but it never gets serialized. On the gRPC path, WorkflowServiceGrpc.CreateSchedule sends the request via RequestMapper.createScheduleRequest(request), there's no setState(...) and t.getState() is never read. So state is dropped and the schedule comes back un-paused, no error.

src/test/java/com/uber/cadence/internal/sync/ScheduleClientImplTest.java
```diff
@@ -343,6 +344,102 @@ public void updateSchedule_cleanTypeOverload_setsFields() throws Exception {
         com.uber.cadence.ScheduleOverlapPolicy.CONCURRENT, req.getPolicies().getOverlapPolicy());
   }
 
+  // --- initialState ---
+
+  @Test
+  public void createSchedule_initialState_pausedWithReasonAndPausedBy() throws Exception {
```
YaweiZhang-930, line 351: These assert ARE before RequestMapper runs — that's why they stay green even though state gets dropped on the way to the proto. Could we add one that runs createScheduleRequest and checks the resulting api.v1.CreateScheduleRequest (getState().getPaused() etc.)? That would've caught this.

abhishekj720:
Superseded by #1083 (IDL bump) and #1082 (feature), split per review feedback.

PR #1082, opened 2026-08-21
feat(schedule): support creating schedule in paused state

abhishekj720:
## What

Adds `ScheduleInitialState` and a five-arg `createSchedule()` overload so callers can create a schedule that starts already paused, without a subsequent `pauseSchedule` call.

## Why

Previously, creating a paused schedule required two round-trips: create, then pause. The server already supports `state` in `CreateScheduleRequest` (added in cadence-idl [#276](https://github.com/uber/cadence-idl/pull/276)), so this wires it end-to-end in the Java client.

## Changes

- `ScheduleInitialState`: new input-only value type (`paused`, `pauseReason`, `pausedBy`). `pausedAt` is intentionally omitted, it is server-populated.
- `ScheduleClient`: new five-arg `createSchedule(scheduleId, spec, action, policies, initialState)` overload. The existing four-arg overload delegates to it with `null`.
- `ScheduleClientImpl`: implementation + `toThriftInitialState` mapper.
- `TypeMapper`: thrift→proto overloads for `schedulePauseInfo` and `scheduleState` (the existing ones were proto→thrift response-only).
- `RequestMapper.createScheduleRequest`: wires `state` into the proto builder so it is not dropped on the gRPC path.

## Test plan

- `ScheduleInitialState`: getters, equals/hashCode, toString, not-equal cases (`ScheduleTypesTest`)
- `createSchedule` wiring: paused with reason+pausedBy, paused with only reason, paused with only pausedBy, paused with no pause info, null initialState, four-arg overload (`ScheduleClientImplTest`)
- `RequestMapper.createScheduleRequest`: verifies state and pauseInfo survive thrift→proto conversion (`ScheduleClientImplTest.createScheduleRequest_stateSerializesToProto`)

### DIFF
--- a/src/main/java/com/uber/cadence/client/ScheduleClient.java
+++ b/src/main/java/com/uber/cadence/client/ScheduleClient.java
@@ -29,6 +29,7 @@ import com.uber.cadence.UpdateScheduleResponse;
 import com.uber.cadence.client.schedule.ScheduleAction;
 import com.uber.cadence.client.schedule.ScheduleCatchUpPolicy;
 import com.uber.cadence.client.schedule.ScheduleDescription;
+import com.uber.cadence.client.schedule.ScheduleInitialState;
 import com.uber.cadence.client.schedule.SchedulePolicies;
 import com.uber.cadence.client.schedule.ScheduleSpec;
 import java.util.List;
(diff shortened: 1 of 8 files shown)


zawadzkidiana (commented):
disregard

YaweiZhang-930 (approved):

dianazawadzki:
> Thanks for this, @abhishekj720 — creating a schedule already paused in one round trip instead of create-then-pause is a nice quality-of-life improvement, and wiring it end-to-end (client type → thrift → proto) rather than stopping at the thrift layer is the right level of completeness.
> 
> Things I liked:
> 
> * `ScheduleInitialState` deliberately omits `pausedAt` since it's server-populated — good call keeping input-only fields out of a type that's only ever used as input.
> * The four-arg `createSchedule` overload delegates to the new five-arg one with `initialState = null`, so existing callers get identical behavior (confirmed by the `createSchedule_fourArg_sendsNoState` test) instead of duplicating request-building logic.
> * Test coverage is thorough on both sides: `ScheduleTypesTest` covers getters/equals/hashCode/toString for the new type, and `ScheduleClientImplTest` covers paused-with-reason-and-pausedBy, paused-with-only-one-of-them, paused-with-neither, null `initialState`, and the thrift→proto serialization round trip. That's a solid set of edge cases for what looks like a small change.
> * `TypeMapper.scheduleState`/`schedulePauseInfo` correctly add the reverse (thrift→proto) direction that was apparently missing — `RequestMapper.createScheduleRequest` was silently dropping `state` on the gRPC path before this.
> 
> Two things worth clarifying before merge:
> 
> 1. **Base branch**: this targets `chore/idl-bump-java`, not `master`, since it depends on the IDL bump adding `state` to `CreateScheduleRequest`. Worth linking that PR here (or in the description) so reviewers/CI know this can't land until the base does — right now it's easy to review this in isolation and forget the dependency ordering.
> 2. **`toThriftInitialState` validation**: if a caller passes `new ScheduleInitialState(false, "some reason", "someone")` (i.e. `paused=false` but with a reason/pausedBy set), the current code silently drops both fields since `pauseInfo` is only attached when `s.isPaused()`. That's probably the right behavior, but is it worth rejecting that combination explicitly (e.g. throwing or at least documenting it), rather than silently discarding data the caller thought they were setting?
> 
> Nicely scoped, well-tested change overall — looks close to mergeable once the base-branch dependency lands.
> 
> _Generated by [Claude Code](https://claude.ai/code)_

reviewing humanly, apologies for claude code trail.

PR #1093, opened 2026-09-04
fix(worker): propagate local activity queue backpressure on full queue

abhishekj720:
## What changed
- `LocalActivityPollTask.apply` now returns `BlockingQueue.offer`'s result instead of always `true`, and restores the interrupt flag on `InterruptedException`.

## Why
`apply` discarded `offer`'s boolean, so when the local-activity queue (capacity 1000) is full the task was dropped but reported as accepted. The caller `ClockDecisionContext.startUnstartedLaTasks` has `if (!applied) return false;` a designed backpressure path (force a new decision task) that was therefore dead code. Result: under queue saturation a task is silently lost and the workflow can wait for a local-activity completion that never arrives.

## How tested
- `./gradlew test --tests "com.uber.cadence.internal.worker.LocalActivityPollTaskTest"` (fails without the fix, passes with it).

## Risks
Low. Restores an intended, previously-dead safety path. The changed return value only differs from prior behavior when the queue is actually full, so the common path is unchanged.

## Note
An earlier revision also tweaked `ClockDecisionContext.awaitTaskCompletion` to honor its timeout; that changed local-activity batching semantics (WorkflowTest LA batching/query failures in CI) and was reverted — the block-until-completion behavior there is intentional.

### DIFF
--- a/src/main/java/com/uber/cadence/internal/worker/LocalActivityPollTask.java
+++ b/src/main/java/com/uber/cadence/internal/worker/LocalActivityPollTask.java
@@ -43,9 +43,13 @@ final class LocalActivityPollTask
   @Override
   public Boolean apply(LocalActivityWorker.Task task, Duration maxWaitAllowed) {
     try {
-      pendingTasks.offer(task, maxWaitAllowed.toMillis(), TimeUnit.MILLISECONDS);
-      return true;
+      // offer returns false when the queue is still full after maxWaitAllowed elapses.
+      // Propagate that instead of dropping the task: the caller relies on a false return
+      // to force a new decision task, otherwise the workflow would wait for a local
+      // activity completion that never arrives.
+      return pendingTasks.offer(task, maxWaitAllowed.toMillis(), TimeUnit.MILLISECONDS);
     } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
       return false;
     }
   }
--- a/src/test/java/com/uber/cadence/internal/worker/LocalActivityPollTaskTest.java
+++ b/src/test/java/com/uber/cadence/internal/worker/LocalActivityPollTaskTest.java
@@ -1,0 +1,67 @@
+/**
+ * Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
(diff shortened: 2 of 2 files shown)


zawadzkidiana (approved):
Hi Abhishek, sorry for the automated claude code review; going through all the comments it left now and reviewing humanly, sorry about that!
--Diana

dianazawadzki:
> Good catch, and a scary one to have sitting silently — `apply` swallowing `offer`'s return value meant `ClockDecisionContext.startUnstartedLaTasks`'s backpressure branch (`if (!applied) return false;`) was dead code, so a full local-activity queue would drop a task while reporting success instead of forcing a new decision task. That's the kind of bug that only shows up under load, and by then a workflow is just silently stuck waiting on a completion that's never coming.
> 
> The fix is minimal and correct: propagate `offer`'s boolean instead of hardcoding `true`, and restore the interrupt flag on `InterruptedException` rather than swallowing it — the latter is easy to overlook but is the right Java concurrency practice here since the caller has no other way to observe that the wait was interrupted.
> 
> Test coverage is a highlight: `applyReturnsFalseWhenQueueIsFull` fills the real bounded queue to capacity and asserts `apply` blocks up to `maxWaitAllowed` before returning `false`, then confirms capacity frees up correctly after a `poll()`. `applyReturnsFalseAndRestoresInterruptWhenInterrupted` separately covers the interrupt-restoration path. Both map directly to the two behaviors this PR fixes.
> 
> One very minor nit, not blocking: the timing assertion `waitedMillis >= 40` against a 50ms `maxWaitAllowed` is a reasonable margin, but timing-based assertions can occasionally be flaky on a loaded CI runner. Not worth blocking on, just flagging in case it ever shows up as an intermittent failure.
> 
> Also appreciated the "Note" in the description calling out that an earlier revision touched `ClockDecisionContext.awaitTaskCompletion` and was reverted after it changed batching semantics — that kind of scope discipline (reverting a tempting-but-unrelated change) makes this much easier to review and reason about in isolation. CI is green across the board. Nice work.
> 
> _Generated by [Claude Code](https://claude.ai/code)_

Hi Abhishek, sorry for the automated claude code review; going through all the comments it left now and reviewing humanly, sorry about that! 
--Diana

